PR #129, opened 2023-05-31
updated dbt core, snowflake, and bigquery versions

### LINKED ISSUE #101: Upgrade to dbt 1.5
I noticed there was a small breaking change in where dbt 1.5 puts its built docs that we will need to respond to, but I don't think there are any major blockers.

britt-allen:
Fixes #101

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -320,33 +320,33 @@ tox = ["tox"]
 
 [[package]]
 name = "dbt-bigquery"
-version = "1.4.3"
+version = "1.5.1"
 description = "The Bigquery adapter plugin for dbt"
 category = "main"
 optional = false
 python-versions = ">=3.7"
 files = [
-    {file = "dbt-bigquery-1.4.3.tar.gz", hash = "sha256:af85c41cdf7efd71eee9a460f5b51d8c59ca903efbd2c6974d1e159c3e871489"},
-    {file = "dbt_bigquery-1.4.3-py3-none-any.whl", hash = "sha256:7f6a7bf40492e68c2782410cf8e118321c4f3844c577d2e7c3c0b79acb307add"},
+    {file = "dbt-bigquery-1.5.1.tar.gz", hash = "sha256:af85c41cdf7efd71eee9a460f5b51d8c59ca903efbd2c6974d1e159c3e871489"},
+    {file = "dbt_bigquery-1.5.1-py3-none-any.whl", hash = "sha256:7f6a7bf40492e68c2782410cf8e118321c4f3844c577d2e7c3c0b79acb307add"},
 ]
 
 [package.dependencies]
 agate = ">=1.6.3,<1.7.0"
-dbt-core = ">=1.4.0,<1.5.0"
+dbt-core = ">=1.5"
 google-cloud-bigquery = ">=3.0,<4.0"
 google-cloud-dataproc = ">=5.0,<6.0"
 google-cloud-storage = ">=2.4,<3.0"
 
 [[package]]
 name = "dbt-core"
-version = "1.4.6"
+version = "1.5"
 description = "With dbt, data analysts and engineers can build analytics the way engineers build applications."
 category = "main"
 optional = false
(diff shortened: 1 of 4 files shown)


britt-allen:
IRL sigh, getting a failure: https://github.com/cagov/data-infrastructure/actions/runs/5137261032

ian-r-rose:
If you add your branch name to the list here you can temporarily build the docs in CI to see if there are any problems: https://github.com/cagov/data-infrastructure/blob/46fb01c401df408d32bd984a7262a70767167c8f/.github/workflows/docs.yml#L5

I had run into a problem where the built docs were landing in a different place so the logic around copying them into our github pages deploy broke. Totally fixable, but annoying. But it's possible that was an unintentional regression that has since been fixed.

ian-r-rose:
@britt-allen are you able to check/fix the docs build?

britt-allen:
I did not realize anything needed fixing since the checks passed. Let me read more carefully what you said above and take a look. Thanks for following up @ian-r-rose

britt-allen:
I cannot update the yaml file bc of current access token settings. Not sure how or where to change those either. I am curious to understand why this PR specifically triggers this change in the docs.yml file? @ian-r-rose

britt-allen:
Docs build complete and ready for merging @ian-r-rose

ian-r-rose:
Looks like the docs build had some problems, cf: https://cagov.github.io/dbt_docs_snowflake/index.html

ian-r-rose:
This is a breaking change to where dbt places the built project in 1.5 notice where it places the docs directory when running 1.4:

https://github.com/cagov/data-infrastructure/actions/runs/5136928618/jobs/9244319759#step:9:66

vs where it places the docs directory when running 1.5:

https://github.com/cagov/data-infrastructure/actions/runs/5149087990/jobs/9271648644#step:9:64

That is what I was referring to in #101. I think it would fix things to just change the `cp` statement.

ian-r-rose:
Found an issue for it: https://github.com/dbt-labs/dbt-core/issues/7465

Looks like it was an unintentional change and will be fixed in the next release of dbt (1.5.2). So it might fix itself if we just wait and then bump the version again.

britt-allen:
So the current docs checks are not capable of catching this issue, correct? And that's why the checks passed despite the build actually being broken? @ian-r-rose

britt-allen:
Should we write a test that doesn't let the check pass if the url is erroring out? I am also having a bit of trouble finding the cp statement you mentioned. May Slack you after a bit more digging. @ian-r-rose

ian-r-rose:
> So the current docs checks are not capable of catching this issue, correct? And that's why the checks passed despite the build actually being broken? @ian-r-rose

Correct, there aren't really any docs checks right now, it's a very manual process

ian-r-rose:
The relevant lines are here: https://github.com/cagov/data-infrastructure/blob/46fb01c401df408d32bd984a7262a70767167c8f/.github/workflows/docs.yml#L47-L54

But it does look like if we wait a week or two this will be fixed with the next dbt release.

ian-r-rose:
Sigh, two steps forward, one step back. This is now blocked by https://github.com/dbt-labs/dbt-core/issues/7937

poetry.lock
britt-allen, line 1: Hey Ian, can I get a tl;dr on why u removed some of the packages you did?

britt-allen (commented):
Can't approve my own PR, but looks good.

poetry.lock
ian-r-rose: These are managed by `poetry`, so I'm not sure. It's a consequence of updating the versions above: somewhere in their dependency tree something required those packages before, and no longer does.

ian-r-rose:
Just testing the docs build on this branch, then I'll revert and if things work well, should be good to go

AeriShan-ODI (approved):

AeriShan-ODI (approved):

AeriShan-ODI:
Tested, no issues found

Issue #119, opened 2023-05-23
Set up production AWS account


ian-r-rose:
Once cagov/Help-Desk#909 is resolved, set up the production account to mirror our dev account.

PR #131, opened 2023-06-12
Modularize AWS terraform config

### LINKED ISSUE #117: Modularize AWS infrastructure
Now that https://github.com/cagov/Help-Desk/issues/909 is nearly complete, it would be good to modularize our AWS configuration to allow for identical dev/prod deploys. We can pull more-or-less the same trick that #100 did, which is to say:

- Refactor our AWS terraform configuration into a module
- Have dev/prod configurations that source the above module.

The painful thing will be migrating state, since the resources will have different names, but I think a relatively simple bash script would do the trick.

ian-r-rose:
Fixes #117 

This does not change any infrastructure, it just moves it into a separate terraform module which can be imported by other modules. I plan to use this to create a production version of our infrastructure (#119).

I *did* have to do some manual terraform state edits, as the names of the resources are changing from `foo` to `module.infra.foo`. The following bash script did the trick:

```bash
for resource in $(terraform state list); do
    terraform state mv ${resource} module.infra.${resource}
done
```


AeriShan-ODI (approved):

AeriShan-ODI (approved):

AeriShan-ODI (approved):

AeriShan-ODI (approved):

AeriShan-ODI (approved):

ian-r-rose:
Thanks for the review!

PR #145, opened 2023-07-20
DOF - Loading census data

britt-allen:


britt-allen:
ready for review @ian-r-rose sorry for the double tag if clicking "ready for review" also notified you

ian-r-rose:
Thanks @britt-allen, I can review this afternoon. In the meantime, there are some linter errors to resolve.

britt-allen:
Yep I am working on those!

jobs/geo/core.py
```diff
@@ -1,44 +1,90 @@
 from __future__ import annotations
```
ian-r-rose, line 1: I recommend renaming this file to `tiger.py`.

```bash
git mv core.py tiger.py
```
to retain history

jobs/geo/core.py
```diff
@@ -1,44 +1,90 @@
...
+    table_names = [
+        "COUNTIES_" + year,
+        "TRACTS_" + year,
+        "BLOCK_GROUPS_" + year,
+        "BLOCKS_" + year,
+        "PLACES_" + year,
+        "PUMAS_" + year,
+        "COUNTY_SUBDIVISIONS_" + year,
+        "PRIMARY_SECONDARY_ROADS_" + year,
+    ]
```
ian-r-rose, line 68: This is quite fragile: if you add anything or reorder the imports then tables will be written under the wrong name in the destination.

I'd recommend instead storing `dfs` in a dictionary, mapping table name to dataframe. You can then iterate over `dfs.items()` to get the name and the frame, and it will no longer be dependent on the order that you happened to write them in.

jobs/geo/core.py
```diff
@@ -1,44 +1,90 @@
...
+            conn,
+            table_name=table_names[x],
+            cluster=False,
+        )
 
 
 if __name__ == "__main__":
     conn = snowflake_connection_from_environment(
-        schema="GEO_REFERENCE",
+        schema="CENSUS_2022",  # TODO: figure out a way to make the year a variable of the user input
```
ian-r-rose, line 81: I'd recommend naming this schema `"TIGER_{YEAR}"` to avoid confusion if we actually load census tables.

If you move the creation of `conn` after you read the year you can use it in the schema name

jobs/geo/core_us.py
```diff
@@ -0,0 +1,115 @@
...
+    coastln["year"] = year
+    dfs.append(coastln)
+
+    # core_based_statistical_area = core_based_statistical_areas(year=year, cache=cache)
+    # core_based_statistical_area['year'] = year
+    # dfs.append(core_based_statistical_area) #avaialble for 2021, but not 2022
+
+    # combined_statistical_area = combined_statistical_areas(year=year, cache=cache)
+    # combined_statistical_area['year'] = year
+    # dfs.append(combined_statistical_area) #avaialble for 2021, but not 2022
```
ian-r-rose, line 53: This would be a good opportunity for a try/except. If `pygris` throws a recognizable error, you can catch it and just skip the table if it's not available for a year.

jobs/geo/core_us.py
```diff
@@ -0,0 +1,115 @@
+from __future__ import annotations
```
ian-r-rose, line 1: I'm not sure I agree with breaking the "US" ones into a separate file. There is enough in common between this and the CA tiger files that I think they could be loaded using the same script.

jobs/geo/core.py
```diff
@@ -1,44 +1,90 @@
...
-    gdf_to_snowflake(
-        gdf,
-        conn,
-        table_name=table,
-        cluster=False,
+    place = places(state=state, year=year, cache=cache)
+    place["year"] = year
+    dfs.append(place)
+
+    puma = pumas(state=state, year=int(year), cache=cache)
```
ian-r-rose, line 45: Why do you need to convert `year` to an `int` here, but not elsewhere?

jobs/geo/core.py
```diff
@@ -1,44 +1,90 @@
...
+        "BLOCKS_" + year,
+        "PLACES_" + year,
+        "PUMAS_" + year,
+        "COUNTY_SUBDIVISIONS_" + year,
+        "PRIMARY_SECONDARY_ROADS_" + year,
+    ]
+
+    for x in range(len(dfs)):
+        gdf_to_snowflake(
+            dfs[x].reset_index(drop=True),
```
ian-r-rose, line 72: I think it's worth a comment here for why you had to `reset_index`

jobs/geo/core.py
ian-r-rose: In fact, we can probably reduce the duplication of the dataframe creation above by creating a mapping from table name to pygris function, then iterating over that to create a mapping from table name to dataframe. Or even doing it in one pass. Pseudocode

```python
loaders = {
  f"TRACTS_{YEAR}": tracts,
  f"BLOCKS_{YEAR}": blocks,
  ... # etc
}
for table_name, loader in loaders.items():
  df = loader(state=state, year=year)
  df["year"] = year
  gdf_to_snowflake(df, conn, ...)
```

jobs/geo/core.py
```diff
@@ -1,44 +1,90 @@
...
+    )
+
+    print(f"Downloading data for CA in year {year}")
+
+    state = "CA"
+    cache = True
+    dfs = []
+
+    county = counties(state=state, year=year, cache=cache)
+    county["year"] = year
```
ian-r-rose, line 26: I'm curious the logic of adding a `year` column to all of these. I don't necessarily object, but I'm not sure it's necessary either.

jobs/geo/core_us.py
britt-allen: Ha I thought of this last night while I was trying to sleep

jobs/geo/core_us.py
britt-allen: I did this to address the `PLR0915 Too many statements` error I was getting. Only other way would be to modify max statements and go against Pep8. Unless you know of a 3rd option? @ian-r-rose

jobs/geo/core.py
britt-allen: I don't recall the exact errors, but basically when running the script it would error out and say can't compare string to int. I thought it was weird it only did that for some but not all as well. When I tested this in jupyter lab it worked without the conversion.

jobs/geo/core.py
britt-allen: Oof let me reproduce the error so I can copy it

jobs/geo/core.py
britt-allen: I just think it's helpful to have another datapoint from a UX POV on what year of census data you are working with. If you think it's truly unnecessary I will remove.

jobs/geo/core_us.py
britt-allen: Perhaps if I remove the line of code that adds the year to the table (that we reference below) that will reduce the statements and address the issue.

jobs/geo/core.py
britt-allen: I knew I could do this with a dictionary, but couldn't remember exactly. Decided to go with what I knew in favor of getting data loaded faster, but glad for the opportunity to refactor and re-learn this method.

jobs/geo/core.py
britt-allen: Thank you for the sample! I knew I could do this with a dictionary, but couldn't remember exactly how. Decided to go with what I knew in favor of getting data loaded faster, but glad for the opportunity to refactor and re-learn this method.

jobs/geo/core_us.py
ian-r-rose: This rule is intended to encourage shorter *functions* rather than files. So you could address the error by breaking it into logical functions. But if you implement some of my other suggestions, I think the function should wind up shorter than it currently is.

Ultimately, I'm not a huge fan of `PLR0915 Too many statements`, so we could also just disable that rule :)

jobs/geo/core.py
britt-allen: done

jobs/geo/core.py
britt-allen: Going to remove for now, can re-add if we think a good idea. Done

jobs/geo/core.py
britt-allen: Done

jobs/geo/core.py
britt-allen: done

jobs/geo/core.py
britt-allen: done

jobs/geo/core_us.py
britt-allen: done

jobs/geo/core_us.py
britt-allen: done

jobs/geo/core.py
britt-allen: done

britt-allen:
Based on this highly technical GIS answer [here](https://gis.stackexchange.com/questions/170839/is-re-projection-needed-from-srid-4326-wgs-84-to-srid-4269-nad-83) I have decided transformation from srid 4269 to srid 4326 is not necessary. If you still believe it to be can you explain why? @ian-r-rose

jobs/geo/tiger.py
```diff
@@ -0,0 +1,103 @@
...
+
+    ca_loaders = {
+        f"COUNTIES_{year}": counties,
+        f"TRACTS_{year}": tracts,
+        f"BLOCK_GROUPS_{year}": block_groups,
+        f"BLOCKS_{year}": blocks,
+        f"PLACES_{year}": places,
+        f"PUMAS_{year}": pumas,
+        f"COUNTY_SUBDIVISIONS_{year}": county_subdivisions,
+        f"PRIMARY_SECONDARY_ROADS_{year}": primary_secondary_roads,
```
ian-r-rose, line 43: I'm open to pushback on this, but I think it might be slightly better to not have the year in the table name, and just have it in the schema name. That would make it easier to write scripts/models that just reference, e.g., `COUNTIES`, and select different years by using the appropriate schema. So there would be less work to do to switch between different years if necessary.

jobs/geo/tiger.py
```diff
@@ -0,0 +1,103 @@
...
+                loader(year=year).reset_index(drop=True),
+                conn,
+                table_name=table_name,
+                cluster=False,
+            )
+    except ValueError:
+        pass
+
+    except DriverError:
+        pass
```
ian-r-rose, line 84: This has a logic problem: because the try/except is *outside* of the loop, the first time we hit a `ValueError` or `DriverError`, the looping will halt, and any subsequent tables *will not be loaded.*

Instead, we'll want to put the try/except *inside* of the loops, so that the looping can continue if a specific table fails.

jobs/geo/tiger.py
```diff
@@ -0,0 +1,103 @@
...
+
+        for table_name, loader in us_loaders.items():
+            gdf_to_snowflake(
+                loader(year=year).reset_index(drop=True),
+                conn,
+                table_name=table_name,
+                cluster=False,
+            )
+    except ValueError:
+        pass
```
ian-r-rose, line 81: It would be good to catch the error, and print it to the console (or log it) so that we are aware of which tables fail to load. Some of them we expect, but others might be a problem.

jobs/geo/tiger.py
```diff
@@ -0,0 +1,103 @@
...
+
+    except DriverError:
+        pass
+
+
+# using .reset_index(drop=True) to address the following UserWarning: Pandas Dataframe has non-standard index of type <class 'pandas.core.indexes.base.Index'> which will not be written. Consider changing the index to pd.RangeIndex(start=0,...,step=1) or call reset_index() to keep index as column(s)
```
ian-r-rose, line 87: This comment has been moved from its context

ian-r-rose (commented):
Thanks @britt-allen, this is looking close!

ian-r-rose:
> Based on this highly technical GIS answer [here](https://gis.stackexchange.com/questions/170839/is-re-projection-needed-from-srid-4326-wgs-84-to-srid-4269-nad-83) I have decided transformation from srid 4269 to srid 4326 is not necessary. If you still believe it to be can you explain why? @ian-r-rose

Let me look into it a bit. Can I ask why you don't want to change to 4326?

britt-allen:
> 

I am not opposed to it. I don't know how to do it (yet). When I started researching a method this was the first thing I came across and it seems as though (from my understanding) that the change is not necessary. Wanted to run it by you though as the GIS expert and get a better understanding of why the change might be crucial if you find it to be.

jobs/geo/tiger.py
britt-allen: oh i did that on purpose so the line wouldn’t be long but will move it back!

ian-r-rose: I'd suggest just breaking it into multiple lines for legibility

jobs/geo/tiger.py
```diff
@@ -0,0 +1,103 @@
...
+        f"CORE_BASED_STATISTICAL_AREAS_{year}": core_based_statistical_areas,
+        f"COMBINED_STATISTICAL_AREAS_{year}": combined_statistical_areas,
+    }
+
+    state = "CA"
+
+    try:
+        for table_name, loader in ca_loaders.items():
+            gdf_to_snowflake(
+                loader(state=state, year=year).reset_index(drop=True),
```
ian-r-rose, line 67: @britt-allen here's an example of how to use geopandas to change the SRID of a dataframe:

```suggestion
                loader(state=state, year=year).reset_index(drop=True).to_crs(epsg=4326),
```

ian-r-rose:
> > 
> 
> I am not opposed to it. I don't know how to do it (yet). When I started researching a method this was the first thing I came across and it seems as though (from my understanding) that the change is not necessary. Wanted to run it by you though as the GIS expert and get a better understanding of why the change might be crucial if you find it to be.

I'm going to do a bit of reading about NAD83 so I don't say something incorrect here. But most datasets are distributed using 4326 (including things like GeoJSON). So my instinct was that it's easier to have that as our default for landing geospatial data in our warehouse.

jobs/geo/tiger.py
britt-allen: good point, agreed!

britt-allen:
@ian-r-rose addressed your latest round of feedback, thank you for your suggestions and opportunities for learning!

jobs/geo/tiger.py
```diff
@@ -0,0 +1,131 @@
...
+            print(
+                f"This ValueError: {value_error} This pertains to this CA loader: {table_name}"
+            )
+            continue
+
+        except DriverError as driver_error:
+            print(
+                f"This DriverError: {driver_error} This pertains to this CA loader: {table_name}"
+            )
+            continue
```
ian-r-rose, line 88: My local linter caught something that I missed! `DriverError` is a subclass of `ValueError`, so this `except` block can never be triggered. It should be safe to remove, and the `ValueError` clause can handle both.

jobs/geo/tiger.py
```diff
@@ -0,0 +1,131 @@
...
+                # index to pd.RangeIndex(start=0,...,step=1) or call reset_index() to keep index as column(s)
+                conn,
+                table_name=table_name,
+                cluster=False,
+            )
+        except ValueError as value_error:
+            print(
+                f"This ValueError: {value_error} This pertains to this CA loader: {table_name}"
+            )
+            continue
```
ian-r-rose, line 82: This `continue` is not necessary (al though it doesn't hurt either)

jobs/geo/tiger.py
```diff
@@ -0,0 +1,131 @@
...
+            print(
+                f"This ValueError: {value_error} This pertains to this US loader: {table_name}"
+            )
+            continue
+
+        except DriverError as driver_error:
+            print(
+                f"This DriverError: {driver_error} This pertains to this US loader: {table_name}"
+            )
+            continue
```
ian-r-rose, line 114: Same as above

ian-r-rose (commented):
Thanks @britt-allen, one last small thing I came across when testing this, and then I think we can merge.

jobs/geo/tiger.py
britt-allen: Thanks for this catch!! I addressed this and everything below so this should be ready for merging now @ian-r-rose

ian-r-rose (approved):
Thanks @britt-allen!

PR #147, opened 2023-07-24
added database details

britt-allen:


AeriShan-ODI (approved):
You can also put these in your dbt profile.yml file. I didn't have this env var set and it worked fine.

ian-r-rose:
> You can also put these in your dbt profile.yml file. I didn't have this env var set and it worked fine.

Yeah, these are more relevant if you are doing work using the terraform provider or Python connector

PR #168, opened 2023-08-11
5 models of building footprints data joined with Tiger Data

britt-allen:
Joined with Tiger Blocks, Block Groups, Tracts, Counties, and County Subdivisions.

Will continue iterating to add docs. 

@Fennisreed can you let me know if there are any other desired columns to. pull in from any of the Tiger tables or any additional Tiger tables you'd like to see joined with Building Footprints data? Docs for Tiger tables are [here](https://www2.census.gov/geo/pdfs/maps-data/data/tiger/tgrshp2022/TGRSHP2022_TechDoc.pdf).


fennisreed:
> Joined with Tiger Blocks, Block Groups, Tracts, Counties, and County Subdivisions.
> 
> Will continue iterating to add docs.
> 
> @fennisreed can you let me know if there are any other desired columns to. pull in from any of the Tiger tables or any additional Tiger tables you'd like to see joined with Building Footprints data? Docs for Tiger tables are [here](https://www2.census.gov/geo/pdfs/maps-data/data/tiger/tgrshp2022/TGRSHP2022_TechDoc.pdf).

Looks good to me - the only additional one that may be useful is incorporated cities from section 4.14.

transform/models/staging/dof_mdsa/stg_footprints_block.sql
```diff
@@ -0,0 +1,13 @@
+with footprints_block as (
+    select 
+f."release",
+f."capture_dates_range",
+f."geometry",
+b."GEOID20",
+b."NAME20"
+from raw_dev.building_footprints.california_building_footprints f
```
ian-r-rose, line 8: This and elsewhere should be a `{{source}}` call

transform/models/staging/dof_mdsa/stg_footprints_block.sql
```diff
@@ -0,0 +1,13 @@
+with footprints_block as (
+    select 
+f."release",
+f."capture_dates_range",
+f."geometry",
+b."GEOID20",
+b."NAME20"
```
ian-r-rose, line 7: Here and below, I'd suggest giving these better names

transform/models/staging/dof_mdsa/stg_footprints_block.sql
```diff
@@ -0,0 +1,13 @@
...
+from raw_dev.building_footprints.california_building_footprints f
+left join raw_dev.tiger_2022.blocks b
+on st_intersects(f."geometry", b."geometry")
+)
+
+select * from footprints_block
```
ian-r-rose, line 13: How computationally expensive is the `block` version of this? If counties took about 90 seconds on a XS, how long did blocks take?

ian-r-rose (commented):
Blocks, block groups, tracts, and counties are all related hierarchically, and the next can be obtained from the prior by truncating the FIPS codes. So I think we only actually *need* to do the `blocks` join, and then the rest could derived from it without doing more expensive spatial joins. This would also consolidate four very similar models into one.

@fennisreed as the resident census data expert, does any of that seem incorrect to you?

ian-r-rose:
> Looks good to me - the only additional one that may be useful is incorporated cities from section 4.14.

For my own understanding, is county subdivisions a superset of incorporated cities?

fennisreed:
> 

From my understanding the county subdivisions represent contiguous statistical areas for minor civil divisions. They can coincide with incorporated cities, but this isn't always the case. To make sure all cities are represented I recommend using the places file.

It should be fine to use the census block FIPS to code all greater levels of census geometry, with the exception of TIGER places. While census geometry remains largely unchanged through the decade, incorporated cities can still have changes to their boundaries on intercensal years.

britt-allen:
Thanks for that clarification Fennis and the guidance Ian. I'll work on
this a bit more and flag my updates for you two!

ian-r-rose:
The naming of the resource directory`dof_mdsa` feels off to me, we're hoping that this dataset will be generally useful, and the fact that it comes to us through an MDSA program doesn't really seem relevant to the actual directory structure.

ian-r-rose:
Our dbt project is still in some flux, but [dbt best practices](https://docs.getdbt.com/guides/best-practices/how-we-structure/2-staging) recommend *not* doing any joins or aggregations in a `stg_` model. In that convention, these would probably be `int_` models.

transform/models/staging/dof_mdsa/stg_footprints_block.sql
britt-allen: Makes sense

transform/models/staging/dof_mdsa/stg_footprints_block.sql
britt-allen: Will update

transform/models/staging/dof_mdsa/stg_footprints_block.sql
britt-allen: Today, Counties took 82 secs, Blocks took 100 secs. Can you help me understand if this information and the difference in speed should change anything I am doing with my code? Blocks is a larger data table (519.7 rows vs 58 rows) so it makes sense to me that the join would take longer and it doesn't seem significantly so despite its size.

britt-allen: Ooh, that was understood from your comments above, but this helps paint the picture another way. Thank you Ian!

britt-allen: It's 519,723 rows

britt-allen:
This is still in progress. I am aware of and addressing the current issues posed in the latest commit.

britt-allen:
Okay, ready for review :)

transform/models/intermediate/geo_reference/int_geo_reference_blocks__mapped.sql
ian-r-rose, line 1: Now this looks like a `stg_` model to me: pure renaming of things, no joins or aggregations.

transform/models/marts/geo_reference/dim_geo_reference_footprints_blocks__joined.sql
ian-r-rose, line 1: I'd be curious to hear @AeriShan-ODI's opinions here. I don't think the dim/fact distinction is particularly useful in this context. It's a geospatial dataset, and while it's probably technically a kimball-ish dim table, I don't really see us joining it with many fact tables. I'd probably just drop the prefix.

transform/models/intermediate/geo_reference/int_geo_reference_places__mapped.sql
ian-r-rose, line 1: Same here

transform/models/marts/geo_reference/_geo_reference__models.yml
```diff
@@ -0,0 +1,88 @@
...
+        description: Current FIPS class definition
+      - name: maf_tiger_feature_class_code
+        description: MAF/TIGER Feature Class Code
+      - name: maf_tiger_feature_class
+        description: >
+          A legal entity incorporated under state law to provide general-purpose
+          governmental services to a concentration of population. Incorporated
+          places are generally designated as a city, borough, municipality, town,
+          village, or, in a few instances, have a different legal description.
+          - and -
```
ian-r-rose, line 83: What's the intended formatting here? Remember that these descriptions are markdown strings.

transform/models/marts/geo_reference/dim_geo_reference_footprints_blocks__joined.sql
```diff
@@ -0,0 +1,42 @@
...
+        blocks."county_fp",
+        blocks."tract_ce",
+        blocks."block_ce",
+        blocks."geo_id",
+        blocks."name",
+        blocks."maf_tiger_feature_class_code",
+        blocks."maf_tiger_feature_class",
+        blocks."urban_rural_area",
+        blocks."housing_unit_count",
+        blocks."pop_count"
```
ian-r-rose, line 37: This is perfectly fine! But it's also an opportunity for the new-ish Snowflake [EXCLUDE](https://docs.snowflake.com/en/sql-reference/sql/select) syntax

I'm a bit concerned that this table is getting a bit wide. To me, the intention here is for the table to be easily filtered and joined with other census data. In that case, I'm not sure that we need things like `maf_tiger_feature_class_code` or `housing_unit_count`.

transform/models/marts/geo_reference/dim_geo_reference_footprints_blocks__joined.sql
```diff
@@ -0,0 +1,42 @@
...
+    from {{ ref('int_geo_reference_blocks__mapped') }}
+),
+
+footprints_and_blocks_joined as (
+    select
+        footprints.*,
+        blocks."county_fp",
+        blocks."tract_ce",
+        blocks."block_ce",
+        blocks."geo_id",
```
ian-r-rose, line 31: For better or worse, I think the canonical name is `GEOID` not "Geo ID"

transform/models/marts/geo_reference/_geo_reference__models.yml
ian-r-rose, line 1: Nice documentation

transform/models/marts/geo_reference/_geo_reference__models.yml
```diff
@@ -0,0 +1,88 @@
...
+  - name: dim_geo_reference_footprints_blocks__joined
+    description: |
+      This data table is a join of the TIGER shapefile, Blocks,
+      with the Microsoft Building Footprints data for the state of CA.
+    columns:
+      - name: release
+        description: The version of the data
+      - name: capture_dates_range
+        description: Each building footprint has a capture date tag from 2019-2020
+      - name: county_fp
```
ian-r-rose, line 13: I don't love `fp` and `ce` as suffixes: I think `fp` stands for FIPS and `CE` stands for Census? Neither seem like particularly useful abbreviations, and are instead probably legacy nonsense from dealing with shapefiles. I think I'd prefer `county_fips`, `tract`, and `block`

transform/macros/map_class_fp.sql
```diff
@@ -0,0 +1,17 @@
+{% macro map_class_fp(class_fp, k, v) -%}
+
```
ian-r-rose, line 2: Clever, do you have a reference to link to for these codes?

transform/macros/map_maf_tiger_feature_class.sql
```diff
@@ -0,0 +1,16 @@
+{% macro map_maf_tiger_feature_class(maf_tiger_feature_class, k, v) -%}
+
+{% set maf_tiger_feature_class_dict = {
+    "G4110" : "Incorporated Place",
+    "G4210" : "Census Designated Place",
+    "G5040" : "Tabulation Block"
```
ian-r-rose, line 6: Are there any other options? Is it worth doing an `ELSE` clause?

transform/models/staging/building_footprints/_building_footprints__models.yml
```diff
@@ -0,0 +1,8 @@
+version: 2
+
+sources:
+  - name: building_footprints
+    database: raw_dev
+    schema: building_footprints
```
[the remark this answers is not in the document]
ian-r-rose: Okay, I've re-run the loading script, and now the data are in `GEO_REFERENCE`. Not sure what happened there.

I'm not *against* a `BUILDING_FOOTPRINTS` schema, but I think we should do it deliberately.

transform/macros/map_class_fp.sql
```diff
@@ -0,0 +1,17 @@
...
+{% set class_fp_dict = {
+    "M2" : "A military or other defense installation entirely within a place",
+    "C1" : "An active incorporated place that does not serve as a county subdivision equivalent",
+    "U1" : "A census designated place with an official federally recognized name",
+    "U2" : "A census designated place without an official federally recognized name"
```
ian-r-rose, line 7: This macro is an interesting pattern as an alternative to a look-up table. I kind of like it, as it essentially moves look up tables into code. I'd be curious of @AeriShan-ODI has any thoughts on this.

ian-r-rose (commented):
This is looking great! The core functionality looks there, most of my comments are around naming conventions and just how granular to get with modeling (I tend to think less granular is good unless we have some shared set of base models)

transform/models/marts/geo_reference/dim_geo_reference_footprints_blocks__joined.sql
ian-r-rose: Thinking about it a bit more: I think we should drop the demographic-related things and just stick to block, tract, etc identifiers here. Things like `pop_count` are kind of confusing here. It looks like it's attached to the building footprint, but maybe it's for the tract, or for the county? What is urban/rural referring to? *We* know that the answer is referring to the block, but it's not clear from a glance at the table.

transform/macros/map_maf_tiger_feature_class.sql
britt-allen: Nope and nope!

transform/macros/map_class_fp.sql
britt-allen: When I was approaching this look up tables came up as an option, but I didn't like the idea of the "clutter" (not sure why I think of it that way, is data hoarding a thing?) of an additional table! I didn't consider fully how this could be a pattern that moves it into code so thank you for recognizing that!

transform/macros/map_class_fp.sql
britt-allen: Yes! I will add it as a comment

transform/models/marts/geo_reference/_geo_reference__models.yml
britt-allen: Makes sense!

transform/models/staging/tiger_2022/stg__blocks.sql
```diff
@@ -0,0 +1,20 @@
...
+        "NAME20" as "name",
+        "MTFCC20" as "maf_tiger_feature_class_code",
+        "UR20" as "urban_rural_area",
+        "HOUSING20" as "housing_unit_count",
+        "POP20" as "pop_count",
+        "geometry"
+    from source
+)
+
+select * from staged
```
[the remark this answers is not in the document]
britt-allen: this is regarding the naming of the cols, right? if so, it's fixed!

transform/models/staging/tiger_2022/stg__places.sql
```diff
@@ -0,0 +1,17 @@
...
+        "PLACENS" as "place_ns",
+        "GEOID" as "geo_id",
+        "NAME" as "name",
+        "CLASSFP" as "class_fp_code",
+        "MTFCC" as "maf_tiger_feature_class_code",
+        "geometry"
+    from source
+)
+
+select * from staged
```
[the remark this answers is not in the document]
britt-allen: this is regarding the naming of the cols, right? if so, it's fixed!

transform/models/intermediate/geo_reference/_int_geo_reference__models.yml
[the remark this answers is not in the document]
britt-allen: I noticed that was done in existing state entity files so I copied the convention.

transform/models/intermediate/geo_reference/int_geo_reference_blocks__mapped.sql
britt-allen: The conventions are a little confusing. When I read dbt's docs on staging, int, and mart models it seemed that renaming things would go into a staging model, transformations (like mapping) would go into int models, and joins would go into mart models. That's what I followed here. But you are recommending this mapping step be moved into the staging model? If so, do I need an int model at all?

transform/models/marts/geo_reference/dim_geo_reference_footprints_blocks__joined.sql
britt-allen: That makes sense, so we should default to dim and fact when we have both, but if there will be no fact tables or vice versa, not use this convention?

transform/models/marts/geo_reference/dim_geo_reference_footprints_blocks__joined.sql
britt-allen: Is the intent to make the table useful for glancing or useful for gathering related dimensions. Can you elaborate on why you think the `maf_tiger_feature_class_code`, `pop_count`, or `housing_unit_count` dimensions are not useful?

transform/models/marts/geo_reference/dim_geo_reference_footprints_blocks__joined.sql
britt-allen: Fixed!

transform/macros/map_class_fp.sql
ian-r-rose: Yeah, another option is to use dbt "seeds", which might be a bit more idiomatic. I don't have a ton of experience with either approach.

transform/models/staging/tiger_2022/stg__blocks.sql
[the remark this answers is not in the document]
ian-r-rose: No, this is in reference to thinking that a pass-through staging model doesn't feel necessary to me. An alternative would be for the joins to just reference the source, and not have a model that doesn't do much.

transform/models/intermediate/geo_reference/int_geo_reference_blocks__mapped.sql
ian-r-rose: > But you are recommending this mapping step be moved into the staging model? If so, do I need an int model at all

Yes, I was thinking that the mapping step feels more like a staging action than an `int` one (but opinions might differ there!). But I am currently feeling like this pipeline isn't sufficiently complex to *need* to be broken into `stg` and `int` in the first place. It could mostly be consolidated into a single model that does the mapping and the join.

To me, `stg` and `int` models start to become worthwhile once you start needing to reuse the same base model in multiple places, but they aren't necessarily needed for their own sake.

transform/models/marts/geo_reference/dim_geo_reference_footprints_blocks__joined.sql
ian-r-rose: I think they have the potential to be misleading: the effective "primary key" for this table is the building footprint. `pop_count` looks like it might refer to the building. So might `housing_unit_count` (what would that mean?). Or whether it's urban or rural.

As the implementer, you know that it's referring to the census *block*, but the casual user of the dataset might not. I think the intended use of this table is to provide *census-related-GEOIDs*, which make it easy to join on any other census data. This goes way beyond `pop_count`, or `housing_unit_count`.

transform/models/marts/geo_reference/_geo_reference__models.yml
ian-r-rose: I meant the dashes around the `and`. Is the intent to emphasize? You have markdown at your disposal, so it could be `*and*`.

transform/models/marts/geo_reference/dim_geo_reference_footprints_blocks__joined.sql
ian-r-rose: Yeah, I think my preference would be to not use `dim` or `fct` prefixes if the dataset doesn't fit that model, and to use it if it does. But I'm curious if @AeriShan-ODI agrees as the resident data modeling expert.

transform/models/intermediate/geo_reference/int_geo_reference_blocks__mapped.sql
britt-allen: That all makes sense. So you're saying I should do my renaming, mapping, and joining in my mart model and the mart model reference the source directly?

transform/models/marts/geo_reference/_geo_reference__models.yml
britt-allen: Oh lol that was because it's two meanings. I guess two sentences is fine and I don't need to structure this way.

transform/models/marts/geo_reference/dim_geo_reference_footprints_blocks__joined.sql
britt-allen: This clarification is helpful, I agree. Thank you!

transform/models/intermediate/geo_reference/int_geo_reference_blocks__mapped.sql
ian-r-rose: Yeah, that's what I'm thinking. If we identify reusable blocks that other models start to use, then it makes more sense to me to start having intermediate models. But it feels premature to me right now.

transform/models/staging/building_footprints/_building_footprints__models.yml
```diff
@@ -0,0 +1,8 @@
+version: 2
+
+sources:
+  - name: building_footprints
+    database: raw_dev
```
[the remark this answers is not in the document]
britt-allen: Thanks for that example! I will skip this since based on our other discussion we don't need a staging table for footprints data and can directly reference it.

transform/models/staging/building_footprints/_building_footprints__models.yml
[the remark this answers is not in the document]
britt-allen: That all makes sense, also going to skip this per our convo about not needed a staging table for footprints data.

transform/models/staging/building_footprints/stg_building__footprints.sql
```diff
@@ -0,0 +1,13 @@
...
+
+staged as (
+    select
+        "release",
+        "capture_dates_range",
+        "geometry"
+    from source
+)
+
+select * from staged
```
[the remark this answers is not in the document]
britt-allen: I agree and good call on breaking away from the dbt training wheels.

transform/models/staging/tiger_2022/stg__blocks.sql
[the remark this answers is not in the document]
britt-allen: Gotchu!

transform/models/staging/building_footprints/_building_footprints__models.yml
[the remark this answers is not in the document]
ian-r-rose: We still need this: whether or not we have a `stg` model, we still need to document our sources in order to refer to them with the `source` macro

transform/models/staging/building_footprints/_building_footprints__models.yml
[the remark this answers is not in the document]
ian-r-rose: As above, we still need to refer to our sources, whether or not the model are broken into staging/intermediate models.

transform/models/marts/geo_reference/dim_geo_reference_footprints_blocks__joined.sql
britt-allen: Makes sense! Would also love Aeri's thoughts.

transform/models/marts/geo_reference/_geo_reference__models.yml
```diff
@@ -0,0 +1,101 @@
...
+          this feature are Count Question Resolution (CQR), current, and
+          tabulation census.
+      - name: urban_rural_area
+        description: >
+          Reserved for 2020 Census urban/rural indicator
+          (2020 Urban Areas are not yet defined)
+      - name: housing_unit_count
+        description: 2020 Census housing unit count
+      - name: pop_count
+        description: 2020 Census population count
```
ian-r-rose, line 58: I don't think we should include these, per our earlier conversation (perhaps also the feature classes).

Update: looks like you've removed some of them from the SQL, but not the documentation

transform/models/marts/geo_reference/geo_reference__blocks_joined.sql
```diff
@@ -0,0 +1,45 @@
...
+),
+
+blocks_renamed as (
+    select
+        "COUNTYFP20" as "county_fips",
+        "TRACTCE20" as "tract",
+        "BLOCKCE20" as "block",
+        "GEOID20" as "geoid",
+        "NAME20" as "name",
+        "UR20" as "urban_rural_area",
```
ian-r-rose, line 20: Let's drop urban-rural-area

transform/models/marts/geo_reference/geo_reference__places_joined.sql
```diff
@@ -0,0 +1,44 @@
...
+        "class_fips_code",
+        {{ map_class_fips("class_fips_code") }} as "class_fips",
+        "geometry"
+    from places_renamed
+),
+
+footprints_and_places_joined as (
+    select
+        footprints.*,
+        places.* exclude "geometry"
```
ian-r-rose, line 39: cool

transform/models/marts/geo_reference/geo_reference__places_joined.sql
```diff
@@ -0,0 +1,44 @@
...
+    select
+        "place_fp",
+        "place_ns",
+        "geoid",
+        "name",
+        "class_fips_code",
+        {{ map_class_fips("class_fips_code") }} as "class_fips",
+        "geometry"
+    from places_renamed
+),
```
ian-r-rose, line 34: Purely stylistic, but here and in the blocks model, I think these two CTEs could be combined. I don't think separating them increases legibility

ian-r-rose (commented):
Looking really close! I noticed a few things

transform/models/marts/geo_reference/_geo_reference__models.yml
```diff
@@ -0,0 +1,101 @@
+version: 2
+
+sources:
+  - name: building_footprints
+    database: raw_dev
```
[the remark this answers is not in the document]
britt-allen: I see, I misunderstood the first time!

transform/models/marts/geo_reference/_geo_reference__models.yml
```diff
@@ -0,0 +1,101 @@
...
+      - name: california_building_footprints
+  - name: tiger_2022
+    database: raw_dev
+    schema: tiger_2022
+    tables:
+      - name: blocks
+      - name: places
+
+models:
+  - name: geo_reference__blocks_joined
```
ian-r-rose, line 17: Thinking about this a bit more: I still think this name needs some work: it doesn't have anything about building footprints, in it, and geo_reference seems not particularly descriptive. Maybe something along the lines of  `geo_reference__building_footprints_with_blocks`? Would be great to get @AeriShan-ODI's opinion here as well

transform/models/marts/geo_reference/_geo_reference__models.yml
britt-allen: when you say perhaps also the feature classes are you referring to `class_fips_code` and `class_fips`?

ian-r-rose: I meant `maf_tiger_feature_class` and `maf_tiger_feature_class_code`. But I'm a bit agnostic there.

transform/models/marts/geo_reference/geo_reference__places_joined.sql
britt-allen: Agree! Fixed

transform/models/marts/geo_reference/_geo_reference__models.yml
britt-allen: Accepting this suggestion for now unless there's further discussion and suggestions to consider.

ian-r-rose (approved):
Thanks @britt-allen!

Issue #169, opened 2023-08-11
DOF – Overlapping footprints code implementaion


britt-allen:
- [x] implement logic for deduplicating footprints
- [x] do performance testing to understand the characteristics of various options
- [x] Possibly refactor logic into reusable macro

jkarpen:
Ian is close on a PR but still needs review. May complete today.

ian-r-rose:
Closing as completed by #190, #175, #178 !

PR #175, opened 2023-08-21
Footprints deduplication

ian-r-rose:
Work towards fixing #169.

This does three-step process for removing duplicate footprints introduced by joining with census blocks when the footprint overlaps more than one census block:

1. Add a surrogate `_tmp_id` to act as a unique identifier for footprints. As long as this isn't treated as a stable identifier or joined with anything else, `seq4()` should be fairly cheap and sufficient. `uuid_string()` would also be fine, and basically impossible to collide with, but I expect that `seq4()` is cheaper as you can more easily do comparisons and (possibly) micropartition logic.
2. For footprints with more than one census block, compute the area of the intersection with the census block. We don't worry too much about SRIDs here, as we are mostly interested in ranking the intersections.
3. Select the footprint with the largest intersection, partitioned/grouped by `_tmp_id`. I'm investigating two approaches, one with window functions and `QUALIFY`, the other with `GROUP BY`. The latter seems provisionally significantly faster.
4. Test that the final number of footprints is identical to the initial.

Still a work in progress, but would appreciate feedback from @britt-allen and @AeriShan-ODI!


transform/models/marts/geo_reference/_geo_reference__models.yml
```diff
@@ -39,6 +39,9 @@ models:
           and the tabulation block number
       - name: geometry
         description: The spatial component of geographic features
+    tests:
+      - count_equals_other:
+          other: source('building_footprints', 'california_building_footprints')
```
britt-allen, line 44: Great to see this logic here! I haven't dove into tests yet so this can be a helpful reference.

transform/models/marts/geo_reference/geo_reference__building_footprints_with_blocks.sql
```diff
@@ -2,7 +2,11 @@ with footprints as (
     select
         "release",
         "capture_dates_range",
-        "geometry"
+        "geometry",
+        -- TODO: is seq4() superior to to uuid_string()? It should be able
+        -- to improve the partitioning below based on micropartition logic.
+        -- but in practice the difference doesn't seem large.
+        seq4() as _tmp_id
```
britt-allen, line 9: Based on what we chatted through live our reasoning about seq4() performing better than uuid_string() makes sense to me and I'm excited to learn the results of your testing.

transform/models/marts/geo_reference/geo_reference__building_footprints_with_blocks.sql
```diff
@@ -24,9 +28,39 @@ blocks as (
...
+        -- TODO: investigate how much savings there are by doing this with
+        -- a `case` statement rather than computing the area for everything.
+        -- also investigate whether there are any savings to be had doing
+        -- this with a self join instead of a window function.
+        case count(*) over (partition by footprints._tmp_id) > 1
+            when false then 1.0
+            when
+                true
+                then st_area(st_intersection(footprints."geometry", blocks."geometry"))
+        end as _tmp_intersection
```
britt-allen, line 41: This is clever to try to do the area calculation when the count is greater than one as opposed to calculating the area for every row. It's unclear to me if / how you'd be able to evaluate if the area calculation is happening sequentially or in parallel, but I suppose one way to ensure that is to split this into two smaller CTEs. Still, I suspect based on our live discussion the alternative using max_by() will be faster.

britt-allen (approved):
I approve so far! Let me know when this is no longer in progress!

transform/models/marts/geo_reference/_geo_reference__models.yml
ian-r-rose: I cribbed this test from one that @AeriShan-ODI wrote in the CalHR repo :)

AeriShan-ODI (approved):
Pretty straightforward...

ian-r-rose:
Dropping a few things I've learned from performance profiling here. There are a few questions I had when writing this:

### Intersections for some or for all?

We only *need* to compute the intersection for footprints which intersect more than one census geometry. Intersection is an expensive operation! So I used a window function to identify the duplicated footprints (`PARTITION BY _tmp_id`), then only computed the intersection for those footprints. But window functions are also expensive! Is it more expensive to do the window function, or unnecessary intersections?


### What kind of temporary ID?

Is it better to use an incrementing integer to track unique footprints or UUIDs? It's probably integers for a couple of reasons:
1. They are cheaper to generate and compare
2. They can be sorted for possible micropartition-based logic.

### Performance considerations for CTEs

If incrementing integers are indeed useful for micropartition-based logic, do we need to add them to a materialized table to effectively "commit" the partitions? Or is the query planner smart enough to also do the same logic if we keep the table-with-temporary-ids as a CTE?

### Deduplicating with window functions vs aggregations

There are (at least) two approaches to deduplicating the footprints once the intersection is calculated. Which is faster?

1. Window function, order by intersection, `QUALIFY row_number() = 1`
2. Group by the id, aggregate by the maximum intersection


I ran some tests across a few different dimensions:

* UUID vs SEQ4
* CTE vs base model for temporary IDs
* Window vs Aggregation for deduplicating
* Window+`CASE` vs intersecting for all

Here are some summary tables (with result times in seconds).

First, investigating CASE + Window vs always computing intersections. In this table, deduplication is always done the same way using aggregation:

  | `CASE` + Window | Always compute intersection
-- | -- | --
UUID in CTE | 326 | 195
UUID in table | 366 | 231
SEQ4 in CTE | 326 | 190
SEQ4 in table | 359 | 222

Second, investigating windowing vs aggregation for deduplicating:

  | Window | Aggregation
-- | -- | --
UUID in CTE | 717 | 195
UUID in table | 441 | 231
SEQ4 in CTE | 703 | 190
SEQ4 in table | 433 | 222

There are a few interesting thing here:

* Using `SEQ4()` is usually slightly faster than using `UUID_STRING()` for temporary IDs. But not so much faster that I think that Snowflake is doing clever things with partitions. It's probably just reflecting the relative cheapness of comparing and sorting integers vs UUIDs.
* Aggregation is *always* faster than windowing for deduplication. This seems to be due to the shuffle that is triggered by windowing. There is *tons* of bytes spilling to disk for the windowing cases, and very little for the aggregation cases. There is a certain sense to this, since shuffling is one of the most expensive things you can do. But in this case, a shuffle shouldn't be necessary, as the footprints effectively come pre-sorted. But Snowflake isn't smart enough to know that they are pre-sorted, so it does the shuffle anyways.  So even though it isn't obvious that windowing would be slower, in practice it seems to be by almost a factor of 2.
* Most of the time keeping the temporary ID generation as a CTE is preferable to pre-generating it as a materialized table. Which makes a certain amount of sense, in that there is likely less I/O in the latter case, and the query optimizer can do more things if the query isn't broken up into separate models. There is a weird exception to this, however! The windowing approach to deduplication has a *HUGE* performance difference when using a CTE vs a pre-materialized table, and the CTE is much worse! I don't know what is going on there. It's not hugely consequential, as aggregation is preferred basically everywhere, but it would be nice to understand it!
* It's not worth it to use a window function to filter for footprints that need intersections. In all cases, it's cheaper to just compute intersections for all footprints. This is disappointing. As above, Snowflake could be smart enough to know that the dataset is already pre-sorted, and avoid an expensive shuffle. But it seems not to be.

## TL;DR

Tentative answers to my questions above:

1. Intersections for all
2. Use SEQ4()
3. It's okay to use CTEs here
4. Aggregation instead of window+QUALIFY

britt-allen:
> 

This is brilliant, this could have all easily lived in your head and in worksheets, but the knowledge transfer and explanation behind choices made is so helpful! Thanks Ian. @jasonlally can we get this on the caldata and/or ODI blog? I'd like for us to post more educational pieces and more of our work in general. That is if @ian-r-rose is okay with that? :)

transform/models/marts/geo_reference/_geo_reference__models.yml
```diff
@@ -39,6 +39,9 @@ models:
           and the tabulation block number
       - name: geometry
         description: The spatial component of geographic features
+    tests:
```
britt-allen, line 42: nice find!

britt-allen (approved):
Such clean work!

ian-r-rose:
Thanks for the review and merge @britt-allen! I do think things like this could be interesting for a blog post (or at least a lunch and learn :) )

jasonlally:
Yeah I think we have a lot of things we can start blogging about. Let me check on process and where most appropriate to post. To clarify, what would the subject be? Would it be on the development of a model to deduplicate geo data using dbt + Snowflake?

ian-r-rose:
> To clarify, what would the subject be? Would it be on the development of a model to deduplicate geo data using dbt + Snowflake? @ian-r-rose

I would probably frame it as a case study in performance optimization of a dbt model in Snowflake on geospatial data. That's where most of the interesting stuff are to me. It would be pretty technical, but there's probably ways to structure things so that different parts are accessible to people at different levels.

britt-allen:
I like that framing Ian. I also like Jason's framing and think performance optimization could be one area of focus as well as what were are doing with abstracting data tables to code with macros and using macros to house other repeatable patterns. I was thinking it could be a long deep dive-y article or a 2-3 part mini series.

PR #178, opened 2023-08-23
Apply deduplication logic to footprints-with-places model

ian-r-rose:
Follow-up to #175, applying identical logic to the places model


britt-allen (approved):
Looks gr8!

PR #182, opened 2023-08-23
Writing building footprints to s3

britt-allen:


jobs/geo/write_building_footprints.py
```diff
@@ -0,0 +1,42 @@
...
+    import shapely
+
+    sql_transform = """
+    alter session set GEOGRAPHY_OUTPUT_FORMAT='WKB';
+    """
+    conn.cursor().execute(sql_transform)
+
+    sql_table = """
+    SELECT *
+    FROM ANALYTICS_DEV.ANALYTICS.GEO_REFERENCE__BUILDING_FOOTPRINTS_WITH_BLOCKS
```
ian-r-rose, line 18: This is likely a big file! Does your laptop struggle with it at all? It might be better to query per county, and process them in serial.

jobs/geo/write_building_footprints.py
```diff
@@ -0,0 +1,42 @@
...
+
+    # write parquet and shape files for every single county locally
+
+    counties = gdf.county_fips.unique()
+
+    for x in counties:
+        gdf[gdf.county_fips == x].to_parquet(f"test_{x}.parquet")
+        gdf[gdf.county_fips == x].to_file(
+            f"test_{x}.shp"
+        )  # this currently outputs .cpg, .dbf, .shx files in addition to .shp, not sure if that is intended
```
ian-r-rose, line 34: I mentioned this on our call: Shapefiles are a multi-file format, the typical way of distributing them is by zipping the folder

jobs/geo/write_building_footprints.py
```diff
@@ -0,0 +1,42 @@
...
+
+from jobs.utils.snowflake import snowflake_connection_from_environment
+
+
+def write_building_footprints(conn):
+    """Grab Microsoft state building footprint data for California from Snowflake and write to an S3 bucket."""
+    import geopandas
+    import shapely
+
+    sql_transform = """
```
ian-r-rose, line 11: This isn't really a transform, it's configuring session parameters

jobs/geo/write_building_footprints.py
britt-allen: I haven't tried yet. I knew the intention was to get the entire dataset so I removed the limit when I was done testing. This feels like a suggestion and not a requirement for the code to work or us to get our desired output, is my understanding correct?

jobs/geo/write_building_footprints.py
britt-allen: I must have missed the part where you mentioned zipping them. I'll look into how to do that. Thank you for the pointer!

jobs/geo/write_building_footprints.py
britt-allen: Will name it better!

jobs/geo/write_building_footprints.py
ian-r-rose: I might expect it to crash your computer. I suggest trying it. Even if it succeeds, I still expect it to work better processing them in serial.

Remember, this dataset is large to the point of being difficult (but not impossible) for a single machine to process, and we've just made it bigger.

jobs/geo/write_building_footprints.py
```diff
@@ -28,10 +31,23 @@ def write_building_footprints(conn):
...
+            os.remove(f"county_fips_{x}.cpg")
+
+            zipped.write(f"county_fips_{x}.dbf")
+            os.remove(f"county_fips_{x}.dbf")
+
+            zipped.write(f"county_fips_{x}.shp")
+            os.remove(f"county_fips_{x}.shp")
+
+            zipped.write(f"county_fips_{x}.shx")
+            os.remove(f"county_fips_{x}.shx")
```
ian-r-rose, line 50: This is a fine way to create a zipfile normally, but OGR supports writing them natively, which is likely less error-prone, and fewer lines of code. Try writing to a filename ending in `.shp.zip`.

jobs/geo/write_building_footprints.py
```diff
@@ -0,0 +1,61 @@
...
+        s3.put(
+            f"footprints_with_blocks_for_county_fips_{county}.parquet",
+            "dof-demographics-dev-us-west-2-public/",
+            recursive=True,
+        )
+        s3.put(
+            f"footprints_with_blocks_for_county_fips_{county}.shp.zip",
+            "dof-demographics-dev-us-west-2-public/",
+            recursive=True,
+        )
```
ian-r-rose, line 53: Since you are using `f"footprints_with_blocks_for_county_fips_{county}"` five times here, it is probably simpler and less error-prone to store it in a variable:

```python
file_prefix = f"footprints_with_blocks_for_county_fips_{county}"
gdf.to_file(f"{file_prefix}.parquet")
gdf.to_file(f"{file_prefix}.shp.zip")
```
etc

jobs/geo/write_building_footprints.py
```diff
@@ -0,0 +1,61 @@
...
+
+        s3 = s3fs.S3FileSystem(anon=False)
+        s3.put(
+            f"footprints_with_blocks_for_county_fips_{county}.parquet",
+            "dof-demographics-dev-us-west-2-public/",
+            recursive=True,
+        )
+        s3.put(
+            f"footprints_with_blocks_for_county_fips_{county}.shp.zip",
+            "dof-demographics-dev-us-west-2-public/",
```
ian-r-rose, line 51: One pattern that I've seen before with datasets that are distributed in multiple formats is to write them to different prefixes. So all the shapefiles could be in `"dof-demographics-dev-us-west-2-public/shp/"` and all of the geoparquet files could be in `"dof-demographics-dev-us-west-2-public/parquet"`. Helps to keep the files a bit more organized in the destination, and allows people to more easily choose which format they want.

jobs/geo/write_building_footprints.py
```diff
@@ -0,0 +1,61 @@
...
+            f"footprints_with_blocks_for_county_fips_{county}.shp.zip",
+            driver="ESRI Shapefile",
+        )
+        print(f"loading footprints_with_blocks_for_county_fips_{county}")
+
+        s3 = s3fs.S3FileSystem(anon=False)
+        s3.put(
+            f"footprints_with_blocks_for_county_fips_{county}.parquet",
+            "dof-demographics-dev-us-west-2-public/",
+            recursive=True,
```
ian-r-rose, line 47: Now that it's a single file rather than a lot of files, `recursive=True` is no longer needed, here and below.

ian-r-rose (commented):
This is looking close!

jobs/geo/write_building_footprints.py
ian-r-rose, line 53: Once you are done copying the file to s3, you and others running the script might appreciate cleaning up the files.

```
os.remove(f"{file_prefix}.parquet")
```
etc

britt-allen: i had already added this after i submitted the pr ;)

britt-allen:
@FennisReed we have building footprint data enriched with TIGER Blocks partitioned by county available as [shape](https://s3.console.aws.amazon.com/s3/buckets/dof-demographics-dev-us-west-2-public?region=us-west-2&prefix=shp/&showversions=false) and [parquet](https://s3.console.aws.amazon.com/s3/buckets/dof-demographics-dev-us-west-2-public?region=us-west-2&prefix=parquet/&showversions=false) files for you to start playing with in public S3 buckets. Please let us know of any feedback you have.

ian-r-rose (approved):
Looks good!

britt-allen:
@Fennisreed those links are to our private AWS console so disregard. I'll get you the right links and some code snippets to work with the data ASAP!

britt-allen:
Hi @Fennisreed,

I hope you are well. As promised last week here are some resources you can use to access shape or parquet building footprint data enriched with TIGER Blocks partitioned by county.  

# **Method 1**
Below are some code snippets you can use to access the files programmatically which is way more efficient given there are 58 shape and parquet files each.

### **Working with a single file**

`import geopandas #this imports the geopandas package for use`

`df_parquet = geopandas.read_parquet("footprints_with_tiger_for_county_fips_001.parquet") # this is reading the parquet file and storing it into a variable
df_parquet.head() #this shows you the head of the data table which by default is the first five rows and their respective column headers`

`df_parquet.shape #this gives you the length (row count) and width (column count) of the data table to understand its size`

`df_parquet.plot(); #this will plot all of the geometries, it may take a minute because of the size`

`df_shape = geopandas.read_file("footprints_with_tiger_for_county_fips_001.shp.zip") #notice the difference in this method versus the one used to read the parquet file
df_shape.head()`

`df_shape.shape`

`df_shape.plot();`

### **Working with multiple files**

`file = open("ca_county_fips_list.txt")` (_this file is attached ⬇️_)
[ca_county_fips_list.txt](https://github.com/cagov/data-infrastructure/files/12551446/ca_county_fips_list.txt)
`lines = file.readlines()`
`file.close()`
`ca_county_fips_list = [line.strip() for line in lines]`
`len(ca_county_fips_list)`

`parquet_dfs = {}`
`shape_dfs = {}`

`for county_fips in ca_county_fips_list[:2]:` (_this is only grabbing the first two counties in the list_)
    `parquet_dfs[f"df_parquet_{county_fips}"] = geopandas.read_parquet(f"s3://dof-demographics-dev-us-west-2-public/parquet/footprints_with_tiger_for_county_fips_{county_fips}.parquet")`
    `shape_dfs[f"df_shape_{county_fips}"] = geopandas.read_file(f"s3://dof-demographics-dev-us-west-2-public/shp/footprints_with_tiger_for_county_fips_{county_fips}.shp.zip")`

I highly recommend commenting out the lines of code for the file type that you will not use. I also do not recommend trying to download all 58 files at once, but instead slicing the ca_county_fips_list like so: `ca_county_fips_list[:5]` this gives you the first five, `ca_county_fips_list[5:10]` then the next five, and so on `ca_county_fips_list[10:15]`. You can chunk it however you'd like!

# **Method 2**
If you want to access the files via your browser download you can do so with this path convention: 
https:// s3.region-code.amazonaws .com/bucket-name/key-name 
- **region code** = us-west-2
- **bucket name** = dof-demographics-dev-us-west-2-public
- **key name - parquet** = parquet/footprints_with_tiger_for_county_fips_001.parquet (_the fips_code number will change_)
- **key name - shape** = shp/footprints_with_tiger_for_county_fips_001.shp.zip (_the fips_code number will change_)

examples: 
- https://s3.us-west-2.amazonaws.com/dof-demographics-dev-us-west-2-public/parquet/footprints_with_tiger_for_county_fips_001.parquet (_clicking this will auto-download the file_)
- https://s3.us-west-2.amazonaws.com/dof-demographics-dev-us-west-2-public/shp/footprints_with_tiger_for_county_fips_001.shp.zip (_clicking this will auto-download the file_)

You can read more about this method and others in [Amazon's docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/access-bucket-intro.html).

Please let me know if you have any questions or run into any issues. We'd love your feedback on the data.

fennisreed:
Thank you for such thorough notes! A few things I ran into:

> `df_shape = geopandas.read_file("footprints_with_tiger_for_county_fips_001.shp.zip") #notice the difference in this method versus the one used to read the parquet file df_shape.head()`

Parquet files load fine, but I receive the following error when trying to access the zipped shapefile. Is this a permissions thing in S3, or perhaps there are additional environments I need to set first? 

`ERROR:fiona._env:AWS_SECRET_ACCESS_KEY and AWS_NO_SIGN_REQUEST configuration options not defined, and /root/.aws/credentials not filled`

Downloading both parquet and shapefile via URL works well. Notably, the zipped shapefiles have duplicate file extensions. For example, _footprints_with_tiger_for_county_fips_003.shp.shp_ can be read by FME but not in an ESRI environment until the additional '.shp' has been removed.

When the shapefile is brought into ESRI it looks like features do not have a coordinate system, and gets placed in the South Atlantic ocean by default. Changing the hosted features to _WGS 1984 Web Mercator (auxiliary sphere)_ would fix the default visualization in both FME and ESRI.

Thank you, Britt!

ian-r-rose:
Hi @fennisreed, thanks for the updates! A few notes:

**Coordinate systems**: Good catch! We had missed setting the CRS on the files, I'm working up a fix right now. You are correct that they are distributed in WGS84, but had the metadata missing.

**S3 permissions**: This is a public S3 bucket, but sometimes S3 clients need a hint that they don't need to sign requests (i.e., provide credentials). There are two ways to handle this:
1. Explicitly tell it you don't want to provide credentials:
    ```python
    import os
    import geopandas

    os.environ["AWS_NO_SIGN_REQUEST"] = "YES"

    gdf = geopandas.read_file("s3://dof-demographics-dev-us-west-2-public/shp/footprints_with_tiger_for_county_fips_003.shp.zip")
    ```
2. Use the HTTP URLs. As publicly accessible files, each of these also has an HTTP URL associated with it (i.e., you could just paste the URL into a web browser to download the files). This should always work without any fuss, but may be a bit slower than using the S3 URI.
    ```python
    import geopandas

    gdf = geopandas.read_file("https://s3.us-west-2.amazonaws.com/dof-demographics-dev-us-west-2-public/shp/footprints_with_tiger_for_county_fips_001.shp.zip")
    ```

**Listing available files**: The CSV that @britt-allen is fine, but if we wind up moving or renaming things (as I suspect will happen a few times), it will probably work better to dynamically list the available files. Is `s3fs` available in your Python environment, or is it possible to install? If so, I'd recommend using it to get the list of files:

```python
import s3fs
fs = s3fs.S3FileSystem(anon=True)  # anon=True is equivalent to AWS_NO_SIGN_REQUEST
fs.glob("s3://dof-demographics-dev-us-west-2-public/**/*.parquet")
```

> Downloading both parquet and shapefile via URL works well. Notably, the zipped shapefiles have duplicate file extensions. For example, footprints_with_tiger_for_county_fips_003.shp.shp can be read by FME but not in an ESRI environment until the additional '.shp' has been removed.

*Shakes fist at Esri*.  GDAL wants the `.shp.zip`, but we can probably work around it.

PR #188, opened 2023-08-25
Add page on dbt performance

AeriShan-ODI:
This is a first pass on discussing general approaches to dbt model performance evaluation and optimization. The premise is less about query optimization, which is already a huge topic, and more about how to deal with the more mundane trade-offs in running models. 

I think there is a lot that could be added here and would love some feedback!


docs/dbt-performance.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Costs
+Although compute time is relatively cheap, it's sometimes possible with larger datasets that need to be frequently refreshed to optimize performance to save enough costs to be worth the time to optimize. In Snowflake you can easily monitor costs in the Admin/Usage section of the Snowflake UI, where you can see credits used by warehouse and role.
+Snowflake also provides several tables with meta information that can be used to derive exact costs for each query - an approach to this, with a ready-use-query can be found in the Select.dev blog post ["Calculating cost per query in Snowflake"](https://select.dev/posts/cost-per-query)
+
+Typically, unless model performance is obviously very poor you are better off adjusting the frequency of runs (end users almost always over-state their desire for data freshness) or reducing data set size either by limiting what you provide or by using incremental models.
```
ian-r-rose, line 11: > end users almost always over-state their desire for data freshness

I feel this in my bones

docs/dbt-performance.md
```diff
@@ -0,0 +1,114 @@
...
+
+### The Tradeoffs
+It is extremely important to balance time spent in optimizing model performance with compute costs and other concerns. If it takes you a day to optimize a model to run only a few seconds faster and save a few pennies per run, it's not likely worth the effort. Similarly, the use of incremental materilization can certainly reduce build time but introduce complexity and require a degree of monitoring to ensure integrity.
+
+## Analyzing Performance in dbt
+With every dbt run or build several artifacts are generated in the target/ directory, including the run_results.json file. This includes detailed information on run execution and many people parse this to create dashboards to report on dbt performance and help with optimization and cost monitoring. There is an important caveate here: simply knowing how long a model took to run is important to uncover which models might need optimization, but cannot tell you anything about why they are performing poorly.
```
ian-r-rose, line 23:
```suggestion
With every dbt run or build several artifacts are generated in the target/ directory, including the run_results.json file. This includes detailed information on run execution and many people parse this to create dashboards to report on dbt performance and help with optimization and cost monitoring. There is an important caveat here: simply knowing how long a model took to run is important to uncover which models might need optimization, but cannot tell you anything about why they are performing poorly.
```

docs/dbt-performance.md
```diff
@@ -0,0 +1,114 @@
...
+For more information refer to the [dbt tags documentation](https://docs.getdbt.com/reference/resource-configs/tags).
+
+
+### 2. Materialization Matters
+
+For a good comparison of materialization options and their trade-offs see the [Materialization Best Practices](https://docs.getdbt.com/guides/best-practices/materializations/2-available-materializations) section of the dbt docs.
+
+**Views**: Are a trade-off between build performance and read/reporting performance. In cases where you are using a BI tool, you should almost always use table materializations unless data storage size is an issue or refresh frequency is so high that cost or scheduling conflicts become a problem. In cases where performance at time of reporting is not an issue (say, you are generating an aggregated report on a monthly basis) then views can be a great way to cut run time. Another case where views can be a good option is with staging data of relatively small size, where your queries are relatively light-weight and you want to ensure fresh data without having to configure separate runs for those models.
+
+### 3. Incremental Models
```
ian-r-rose, line 73: This feels like it should be at the same level of organization as **Views** (i.e., under the "Materialization Matters" subhead)

docs/dbt-performance.md
ian-r-rose: You use the term "incremental models" a few times before defining it below. I don't think that's a huge problem, but perhaps it's worth a parenthetical and a link to the below anchor tag?

ian-r-rose (commented):
This is great! I don't have any major suggestions, I think it's all valuable content.

I agree that in-depth query optimization strategies are probably out of scope for this page, so this is just me musing about follow-up content: I think that there could be some value in a rough ranking of "How expensive are various SQL operations likely to be?". In my experience, having a decent intuition of "this is likely to be cheap" vs "this is likely to be expensive" is really valuable.

Off the top of my head, from most expensive to least expensive:

1. Many-to-many joins
2. Window functions partitioned on unsorted data
3. Few-to-many joins
4. Window functions partitioned on sorted data
5. Sorting/ordering
6. Groupby/aggregations on unsorted group keys
7. Groupby/aggregations on sorted group keys
8. Unit conversions, renaming, parsing

docs/dbt-performance.md
AeriShan-ODI: Yeah, that makes sense!

britt-allen (approved):
This looks good Aeri, thank you for putting it together.

docs/dbt-performance.md
ian-r-rose: @AeriShan-ODI do you agree with this, or want to go in as is?

ian-r-rose (approved):
Thanks @AeriShan-ODI!

PR #189, opened 2023-08-29
renamed cols, removed null data

britt-allen:
needed to remove null data so i can finish with my data writing script. i plotted the data where county_fips and place_fips was null and it does not look to be in california or useful thus prompting the removal of it.

### DIFF
--- a/transform/models/marts/geo_reference/_geo_reference__models.yml
+++ b/transform/models/marts/geo_reference/_geo_reference__models.yml
@@ -39,9 +39,6 @@ models:
           and the tabulation block number
       - name: geometry
         description: The spatial component of geographic features
-    tests:
-      - dbt_utils.equal_rowcount:
-          compare_model: source('building_footprints', 'california_building_footprints')
   - name: geo_reference__building_footprints_with_places
     description: |
       This data table is a join of the TIGER shapefile
@@ -71,6 +68,3 @@ models:
         description: Current FIPS class definition
       - name: geometry
         description: The spatial component of geographic features
-    tests:
-      - dbt_utils.equal_rowcount:
-          compare_model: source('building_footprints', 'california_building_footprints')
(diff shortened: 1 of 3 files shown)


britt-allen:
ready for review @ian-r-rose

ian-r-rose:
How many footprints do not match a block? Do you have a map showing where they are?

An alternative would be to just discard them in the partition-then-write-to-s3 step, rather than in the dbt model.

britt-allen:
I considered that, but am curious why you are recommending removing them in that step rather than in the model? @ian-r-rose 

20,595 for Blocks
1,190,757 for Places

That's the "map" of the 20,595 polygons plotted for Blocks
Screenshot 2023-08-29 at 3 37 12 PM

ian-r-rose:
> I considered that, but am curious why you are recommending removing them in that step rather than in the model?

Well, with "places", it makes perfect sense that there would be lots of buildings that are not in a census-defined-place (e.g., it could be in remote communities, or infrastructural buildings). I think I would want to be able to see those, and decide whether I want them based on the needs of the analysis. Note that about 10% of all footprints don't match a "place"!

Blocks are a bit trickier. In theory, they should cover the whole state. I'd like to know a bit more about those blocks which don't match. 20K is still a lot of buildings (roughly a fifth of a percent). It looks like a lot of them are on the Mexican and Arizona borders. I might recommend using `gdf.explore()` to get a better sense of which ones don't match a block and why.

As to why I would prefer doing the filtering later: I like having the data integrity check that the row count is unchanged, and would be sad to lose it. From a perspective of the python script, it's quite easy to not write a file for the `None` county fips code.

ian-r-rose:
Put another way: you are effectively manually changing the "left join" into an "inner join". But we deliberately chose a "left" join to retain footprints that don't necessarily match.

britt-allen:
That all makes sense. I shall close this branch then and do some more exploring and add the code to filter out nulls the load script.

ian-r-rose:
I'm still quite curious to learn more about the footprints that don't match a block, I think it merits further thought!

britt-allen:
That was the plan, just closed the ticket. :)

PR #190, opened 2023-09-01
Refactor geospatial dedupe

ian-r-rose:
This refactors the geospatial deduplication logic from #175 and #178 into a reusable macro. It also consolidates the "blocks" and "places" model into a single model, which uses the macro twice.

Marking as a draft for now because I'll want to update a couple of things once #182 is in, but this should be ready for some review.


transform/models/marts/geo_reference/geo_reference__building_footprints_with_tiger.sql
```diff
@@ -0,0 +1,80 @@
...
+
+footprints_with_blocks as (
+    {{ spatial_join_with_deduplication(
+       "footprints",
+       "blocks",
+       footprint_cols,
+       block_cols.values(),
+       left_geom='"geometry"',
+       right_geom='"geometry"',
+       kind="inner",
```
ian-r-rose, line 62: @britt-allen here is that inner join!

transform/models/marts/geo_reference/_geo_reference__models.yml
```diff
@@ -29,39 +29,19 @@ models:
...
           Census county FIPS code, 2020 Census tract code, and 2020 Census block number
-      - name: name
-        description: >
-          2020 Census tabulation block name; a concatenation of 'Block'
-          and the tabulation block number
-      - name: geometry
-        description: The spatial component of geographic features
-    tests:
-      - dbt_utils.equal_rowcount:
-          compare_model: source('building_footprints', 'california_building_footprints')
```
ian-r-rose, line 44: A shame to lose this test, but with the inner join, the number of footprints is no longer conserved! One approach to restore it would be to do the "in-California" filtering in an intermediate model, then compare the row-count with that.

transform/models/marts/geo_reference/geo_reference__building_footprints_with_tiger.sql
```diff
@@ -0,0 +1,80 @@
+/* Columns selected and renamed from footprints, blocks, and places
+are reused in several places. To avoid repeating them with the possibility
+for errors, we set them in lists/dictionaries where appropriate. */
+{%- set footprint_cols = ['"release"', '"capture_dates_range"'] -%}
+
+{%- set block_cols = {
+      '"COUNTYFP20"': '"county_fips"',
+      '"TRACTCE20"': '"tract"',
+      '"BLOCKCE20"': '"block"',
+      '"GEOID20"': '"block_geoid"',
```
ian-r-rose, line 10: @AeriShan-ODI I'm curious if you have patterns for this that you like: I didn't like duplicating the list of columns in several places as it's annoying to keep them in sync if you move or rename things. But maybe this is too-much-clever-Jjinja?

jobs/geo/write_building_footprints.py
```diff
@@ -38,7 +38,7 @@ def write_building_footprints(conn):
 
         gdf = gdf[gdf.geometry.geom_type != "GeometryCollection"]
 
-        file_prefix = f"footprints_with_blocks_for_county_fips_{county}"
+        file_prefix = f"footprints_with_tiger_for_county_fips_{county}"
```
ian-r-rose, line 41: What do you think about this name, since the files now include county, tract, block, and place data?

britt-allen:
Is this still draft now that #182 is in?

jobs/geo/write_building_footprints.py
britt-allen: That makes sense to me as a name change!

transform/models/marts/geo_reference/_geo_reference__models.yml
britt-allen: We can make that a to do!

transform/macros/map_class_fp.sql
```diff
@@ -12,10 +12,10 @@
 } -%}
 
 case
-    {% for k, v in class_fips_dict.items() %}
+    {% for k, v in class_fips_dict.items() -%}
```
britt-allen, line 15: what's this dash here and below for?

transform/macros/_macros.yml
britt-allen, line 1: I didn't realized macros needed YAML-style documentation. This is advanced stuff for me, thanks for modeling this!

ian-r-rose:
> Is this still draft now that https://github.com/cagov/data-infrastructure/pull/182 is in?

Yeah, I'm still validating one or two things. Probably ready on Tuesday!

transform/macros/_macros.yml
ian-r-rose: They don't need it, but documenting it makes it show up in the docs

transform/macros/map_class_fp.sql
ian-r-rose: This trims some whitespace in the rendered file. Entirely cosmetic

transform/macros/spatial_join_with_deduplication.sql
britt-allen, line 1: impressive!

transform/models/marts/geo_reference/geo_reference__building_footprints_with_tiger.sql
britt-allen, line 1: this is an abstract painting of code 🤩

ian-r-rose: haha, I perhaps went a bit overboard with my jinja

transform/models/marts/geo_reference/geo_reference__building_footprints_with_tiger.sql
britt-allen: what's making you feel like it may be too much too clever? and what's the alternative (if not going back to doing the join and dedupe in the marts models themselves)

britt-allen (approved):
This is all such complex work. Thank you for breaking it down and explaining it step by step with helpful comments and descriptions.

transform/models/marts/geo_reference/geo_reference__building_footprints_with_tiger.sql
ian-r-rose: An alternative would be to just list out the column names every time I need them. It would be more repetition, but perhaps a bit easier to read. This is somewhat independent of whether the dedupe macro is a good idea

AeriShan-ODI: Caveat: I don't know or really understand geospatial data. 

TBH, I do feel like it's "too-much-clever-Jjinja". I mean, it certainly is clever... but in these situations I always try to curb my own desire to be clever with how that gets looked at and maintained down the road. For example, it wasn't immediately obvious what you meant by deduplication - deduping records or columns (seems like both). The sort of abstraction that happened in the macro makes it hard to understand the operation and purpose. 
So with the caveat that I think you should keep this for the cleverness-as-interesting-and-educational factor, but with perhaps more descriptive documentation, my own approach in this would likely be to simply spell out the column names and use SELECT... EXCLUDE... or use a macro to only handle column name collisions. 

All this said, I'm a bit on the fence - we have the opportunity to push boundaries and be clever so why not? I'm so used to situations where cleverness ends up costing money

ian-r-rose: > Caveat: I don't know or really understand geospatial data.

I'm a proponent of "spatial isn't special", i.e, we should use the same set of tools for working with spatial data, it's just another data type with some functions that know how to operate on it. All of which is to say, I think any code style and performance notes you have are fair game!

Here I'm writing this jinja-inflected SQL as if it were Python, and it probably doesn't help the legibility for people who are expecting SQL. I'll change it to just list the column names, there is probably enough jinja-weirdness going around that this is a bridge too far.

britt-allen: I appreciate this discussion and the conclusion reached!

AeriShan-ODI (approved):

britt-allen:
What's the latest on this PR? @ian-r-rose

ian-r-rose:
> What's the latest on this PR? @ian-r-rose

Some counts weren't quite what I expected, so I kept it as a draft so I could investigate further. I should be able to finish it this afternoon!

ian-r-rose:
Okay, this is finally ready! [0873113](https://github.com/cagov/data-infrastructure/pull/190/commits/0873113f5be4a46e9c06ff75e65873da4b2f8e8b) was a tricky bug that took me too long to track down, but was resulting in some unexpected nulls at the ~10% level.

britt-allen (approved):

PR #194, opened 2023-09-06
Simplify Snowflake auth in CI

ian-r-rose:
Now that we are on dbt>1.5, we should be able to simplify private key handling in CI


ian-r-rose:
Ran into some issues with PEM format, closing for now and will check again in another point release

PR #202, opened 2023-09-11
Correct mart table name

ian-r-rose:
Follow-up to #190


britt-allen (approved):
thanks for the correction Ian!

PR #205, opened 2023-09-19
Set crs on footprints

ian-r-rose:
Some fixes for https://github.com/cagov/data-infrastructure/pull/182#issuecomment-1725901699

* Set the CRS on datasets before writing them to S3
* Ensure the zipped shapefiles just end in `.zip` rather than `.shp.zip`, which was causing some problems with Esri tooling.


britt-allen (approved):
Thanks for these changes and the helpful comments, Ian!

PR #207, opened 2023-09-19
Global ml building footprints

### LINKED ISSUE #177: Load new global buildings dataset to Snowflake
From @ian-r-rose :

> Following up on our discussion yesterday about the frequency of updates for the US building footprints: I came across [this](https://blogs.bing.com/maps/2023-06/Bing-Maps-Global-Building-Footprints-released) while poking about the Planetary Computer datasets yesterday. It's basically the same idea as the US building footprints, except:
> - global
> - more recently updated
> - distributed in a slightly more parallel-computing-friendly format
> It includes the US, and a large part of the rest of the world. Unfortunately, they aren't partitioned by state, but instead by Bing tile quadkeys. But scoping them down to California wouldn't be a huge amount of work.
> 
> They don't go quite so far as to say that they are deprecating the US-specific dataset, but I wonder if they are headed in that direction. This certainly looks like where Microsoft is putting more energy right now.
> 
> I don't think it would be a huge lift to load both of the datasets for comparison, so we don't have to make a call on this right now, but it's worth thinking about.

From Fennis:

> I'm enamored with this dataset. It looks like its the same methodology, just with a wider scope. The [Chicago example even has a follow through comparison](https://github.com/microsoft/GlobalMLBuildingFootprints/issues/36).
> 
> My gut reaction is that this is more recent and would serve us better that the US-specific dataset anyway. Especially if the update frequency is improved.

ian-r-rose:
Fixes #177 

Still a work in progress, but opening for visibility


ian-r-rose:
Okay, this is ready for some review

airflow/dags/geo_reference/load_building_footprints.py
```diff
@@ -22,11 +22,11 @@ def building_footprints_dag():
     """DAG for loading MS Building footprints dataset."""
     submit_batch_job = BatchOperator(
         task_id="load_footprints",
-        job_name="california_building_footprints",
+        job_name="us_building_footprints",
```
ian-r-rose, line 25: @britt-allen with this PR there are two different building footprints datasets. I've tried to go through and use consistent naming for them, following the actual Microsoft repos. The older one is called `us_building_footprints`, and the newer one is called `global_ml_building_footprints`

jobs/geo/data/california.geojson
ian-r-rose, line 1: I found it useful to distributed a geojson of the shape of california with this repo. It's small enough that I don't think it's a huge problem

jobs/geo/load_global_ml_building_footprints.py
```diff
@@ -0,0 +1,75 @@
...
+            table_name="GLOBAL_ML_BUILDING_FOOTPRINTS",
+            overwrite=overwrite,
+            strict_geometries=False,
+        )
+        overwrite = False  # For all subsequent gdfs, append
+
+
+if __name__ == "__main__":
+    conn = snowflake_connection_from_environment(
+        schema="BUILDING_FOOTPRINTS",
```
ian-r-rose, line 72: In the past we've gone back and forth between storing this in `RAW.BUILDING_FOOTPRINTS` and `RAW.GEO_REFERENCE`. Now that we have two of them, I'm leaning back towards using a `BUILDING_FOOTPRINTS` schema for these. Thoughts?

jobs/utils/snowflake.py
```diff
@@ -197,20 +207,34 @@ def gdf_to_snowflake(
         for c, dtype in gdf.dtypes.to_dict().items():
             if type(dtype) == geopandas.array.GeometryDtype:
                 if epsg == WGS84:
-                    cols.append(f'TO_GEOGRAPHY("{c}") AS "{c}"')
+                    cols.append(
+                        f'{"" if strict_geometries else "TRY_"}TO_GEOGRAPHY("{c}") AS "{c}"'
```
ian-r-rose, line 211: There were some invalid building footprints in the dataset, no matter how hard I tried to make them valid. So I think it makes sense, in the spirit of ELT, to go ahead and allow the invalid ones through, cast them using [`TRY_TO_GEOGRAPHY`](https://docs.snowflake.com/en/sql-reference/functions/try_to_geography), and filtering out nulls later.

jobs/utils/snowflake.py
```diff
@@ -197,20 +207,34 @@ def gdf_to_snowflake(
...
+            )
+        else:
+            if cluster:
+                raise ValueError(
+                    "Clustering on existing table (overwrite=False) not supported"
+                )
+            sql = (
+                f'''INSERT INTO "{database}"."{schema}"."{table_name}"'''
+                f'''\nSELECT \n{",".join(cols)} \nFROM "{database}"."{schema}"."{tmp_table}"'''
+            )
```
ian-r-rose, line 237: In the past we were overwriting every time. But with this dataset being split into different quadkeys, we have to have logic for appending to an existing table.

transform/models/marts/geo_reference/geo_reference__global_ml_building_footprints_with_tiger.sql
ian-r-rose, line 1: Kind of neat to see some of the deduplication logic paying off more! This is almost identical to the existing one.

jobs/geo/write_building_footprints.py
```diff
@@ -51,21 +55,30 @@ def write_building_footprints(conn):
         s3 = s3fs.S3FileSystem(anon=False)
         s3.put(
             f"{file_prefix}.parquet",
-            "s3://dof-demographics-dev-us-west-2-public/parquet/",
+            f"s3://dof-demographics-dev-us-west-2-public/{kind}_building_footprints/parquet/{file_prefix}.parquet",
```
ian-r-rose, line 58: Here is a proposed change to the naming structure in S3. Now that there are two datasets, it felt weird to me to have `parquet/` or `shp/` be the top-level folders. It now looks like
```
global_ml_building_footprints/parquet/county_fips_001.parquet
global_ml_building_footprints/shp/county_fips_001.zip
us_building_footprints/parquet/county_fips_001.parquet
us_building_footprints/shp/county_fips_001.zip
```

airflow/dags/geo_reference/load_building_footprints.py
britt-allen: thank you for updating the names to be more consistent with Microsoft and more descriptive

jobs/geo/load_global_ml_building_footprints.py
britt-allen: I agree, a Building Footprints schema seems like the best path going forward. Shall we delete RAW.GEO_REFERENCE. I worry about accumulating resources we are not using and cluttering our warehouse.

jobs/utils/snowflake.py
britt-allen: That makes sense to me, thank you for trying and for explaining your final decision.

jobs/utils/snowflake.py
britt-allen: By "in the past" are you reference to the US Building Footprints data?

transform/models/marts/geo_reference/geo_reference__global_ml_building_footprints_with_tiger.sql
britt-allen: That is neat! Clean and efficient!

jobs/geo/write_building_footprints.py
britt-allen: To clarify, the top-level folders will now be either US-based or Global-based with both zip and parquet files stored in either?

jobs/geo/load_global_ml_building_footprints.py
ian-r-rose: Yeah, when this is in, I think we can go ahead and delete `GEO_REFERENCE`.

jobs/utils/snowflake.py
ian-r-rose: Yes, that and other single-file geospatial datasets we've loaded (e.g., all the tiger ones)

jobs/geo/write_building_footprints.py
ian-r-rose: Correct. It's in there now if you want to take a look

jobs/geo/data/california.geojson
britt-allen: Why do you find it useful? What use case(s) would it serve?

jobs/geo/write_building_footprints.py
```diff
@@ -16,19 +16,23 @@ def write_building_footprints(conn):
     """
     conn.cursor().execute(sql_alter)
 
-    counties = """
+    ref = (
```
britt-allen, line 19: this is clever

jobs/geo/write_building_footprints.py
```diff
@@ -39,7 +43,7 @@ def write_building_footprints(conn):
 
         gdf = gdf[gdf.geometry.geom_type != "GeometryCollection"]
 
-        file_prefix = f"footprints_with_tiger_for_county_fips_{county}"
+        file_prefix = f"county_fips_{county}"
```
britt-allen, line 46: nice clean up

britt-allen (approved):
Thanks Ian!

jobs/utils/snowflake.py
britt-allen: Got it and we'll still overwrite for those, but append for the Global dataset?

jobs/geo/data/california.geojson
ian-r-rose: This is used for figuring out which quadkeys intersect California (in conjunction with the `mercantile` library)

jobs/utils/snowflake.py
ian-r-rose: Correct, it's controlled using the `overwrite` parameter.

PR #208, opened 2023-09-20
Codespaces proof-of-concept

### LINKED ISSUE #203: Investigate GitHub codespaces as a serverless hosted coding environment
GitHub codespaces can be configured to spin up ephemeral development environments, including using VS Code and JupyterLab. This could be a useful thing to deploy for people who are uncomfortable setting up python environments themselves.

This issue is to do a bit of experimentation with codespaces to see how well it would work for this use case, including:

* dbt workflows
* python GIS workflows
* docs authoring (?)

If it works well, it could be a nice thing to include in our infrastructure template.

ian-r-rose:
Fixes #203

Would appreciate feedback from @britt-allen or @matthewzhou. There is some documentation for how to set things up [here](https://github.com/cagov/data-infrastructure/blob/devcontainer/docs/codespaces.md)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,32 @@
...
+    "secrets": {
+        "SNOWFLAKE_USER": {
+            "description": "Your Snowflake Username"
+        },
+        "SNOWFLAKE_PASSWORD": {
+            "description": "Your Snowflake Password"
+        },
+        "DBT_SCHEMA": {
+            "description": "The dev schema into which to build your dbt models (e.g. DBT_<YOURNAME>)"
+        }
```
ian-r-rose, line 29: These are the secrets that you need to set in order to do dbt-related work in the codespace

.pre-commit-config.yaml
```diff
@@ -84,8 +84,8 @@ repos:
     hooks:
       - name: Dbt deps
         id: dbt-deps
-        language: python
-        entry: dbt deps --project-dir=transform
+        language: system
+        entry: poetry run dbt deps --project-dir transform
```
ian-r-rose, line 88: Caught some real bugs in the pre-commit set-up when doing this!

mkdocs.yml
```diff
@@ -27,16 +27,18 @@ markdown_extensions:
...
-  - dbt Performance: dbt-performance.md
   - Snowflake: snowflake.md
-  - Writing Documentation: writing-documentation.md
-  - dbt Snowflake Project: dbt_docs_snowflake/index.html
-  - dbt BigQuery Project: dbt_docs_bigquery/index.html
+  - dbt:
+      - Overview: dbt.md
+      - dbt Performance: dbt-performance.md
+      - dbt Snowflake Project: dbt_docs_snowflake/index.html
+      - dbt BigQuery Project: dbt_docs_bigquery/index.html
```
ian-r-rose, line 41: Drive-by change, organizing some of the dbt-related docs into a folder

britt-allen:
There are no instructions for what to do after you get to this screen. This screen comes immediately after clicking "New with options..." @ian-r-rose 
Screenshot 2023-10-31 at 4 42 08 PM

britt-allen:
I did not have an option to add "Recommended Secrets" as the [docs](https://github.com/cagov/data-infrastructure/blob/devcontainer/docs/codespaces.md) say immediately following creation of a codespace with options. This is the screen (below) I was taken to immediately after. It's unclear if I should expect to see the form listed in the docs or if I have to navigate to it. If the latter the instructions are lacking. 
Screenshot 2023-10-31 at 4 45 46 PM

ian-r-rose:
Did you try to create the codespace from the `devcontainer` branch?

ian-r-rose:
You should go to the `devcontainer` branch to create the codespace. If this is in `main`, that step will not be necessary, so it seemed incorrect to add it to the docs

britt-allen:
Wow this was SO easy, I am not sure whatever magic you did in the background, but this just works. Thanks @ian-r-rose. I cannot wait to show this off to future MDSA trainees. 

One question the docs doesn't answer: "_Should we let Visual Studio Code periodically run `git fetch`?_" this is a pop up I got while using codespaces. Want to provide guidance, but seems like it's a personal preference thing. If I do have `git fetch` running in the background can I still run `git pull` to get changes into my working directory or will I only be able to use `git merge` since `git fetch` is being run?

britt-allen (approved):

ian-r-rose:
Thanks for testing it out!

> One question the docs doesn't answer: "Should we let Visual Studio Code periodically run git fetch?" this is a pop up I got while using codespaces. Want to provide guidance, but seems like it's a personal preference thing. 

To be honest, I'm not sure what to recommend here, I don't really understand what the VS Code git extension does when you say yes. I suspect you are right that it's a personal preference thing, though I usually avoid automated git actions because (in my experience) it's an easy way to get into trouble with your git history.

> If I do have git fetch running in the background can I still run git pull to get changes into my working directory or will I only be able to use git merge since git fetch is being run?

I think you should still be able to run normal git operations, even if the extension is also doing things occasionally in the background. But I'm not 100% sure, having not really experimented with it myself.

PR #213, opened 2023-10-03
Include area in square meters to the building footprint datasets

### LINKED ISSUE #212: DOF - add footprint area
Adding a new column to calculate footprint area and then adding that to the public s3 files

ian-r-rose:
Fixes #212


britt-allen (approved):
Looks great! Thank you for enriching the data and adding these docs, Ian.

PR #228, opened 2023-10-25
Add logging user in preparation for ingesting logs into sentinel

### LINKED ISSUE #225: Create logging service user and warehouse
In preparation for #196, we need to create a logging service user and warehouse with permissions on the "Snowflake" database, following [these](https://learn.microsoft.com/en-us/azure/sentinel/data-connectors/snowflake-using-azure-function) instructions. This should be fairly doable using Terraform.

ian-r-rose:
Fixes #225


terraform/snowflake/environments/prd/main.tf
AeriShan-ODI, line 1: So are we using credentials for account admin here? If so, how are they provided? Stored?

ian-r-rose: The local credentials (probably your username and key-pair) would need to have access to `ACCOUNTADMIN`. Previously, they just needed access to `SECURITYADMIN`. It's unfortunate, but I don't see a way around it: providing access to the `SNOWFLAKE` database needs `ACCOUNTADMIN` as far as I know.

Right now, `terraform apply` still needs to be run by a user locally: if we move this to a bot-driven CD, we'd need to have that bot also have `ACCOUNTADMIN` privileges, and store the credentials in a secure way.

ian-r-rose:
Thanks for the review!

terraform/snowflake/environments/prd/main.tf
AeriShan-ODI: Thanks for the clarification!

PR #233, opened 2023-10-30
Schedule global ml footprints dag

### LINKED ISSUE #224: Run GlobalML building footprints pipeline on a schedule
We are currently running a scheduled Airflow job for the US building footprints dataset, but not the more-frequently-updated GlobalML building footprints dataset.

ian-r-rose:
Work in progress. Fixes #224


ian-r-rose:
I'm running into some troubles here in the AWS Batch jobs that I can't reproduce locally -- will update with details soon

terraform/aws/modules/infra/network.tf
```diff
@@ -21,9 +21,9 @@ resource "aws_security_group" "batch" {
 
   egress {
     description = "Allow ECS tasks to talk to the internet"
-    from_port   = 443
-    to_port     = 443
-    protocol    = "tcp"
+    from_port   = 0
+    to_port     = 0
+    protocol    = "-1"
```
ian-r-rose, line 26: So here is why this branch took so long to work out: the security group rule for ECS tasks had been configured to allow egress over HTTPS only (that's port 443). At some point this summer, the `snowflake-connector-python` package was updated, and in those updates it started making a few exploratory connections over HTTP (port 80). I think these connections are just sniffing out Snowflake connectivity and features, but to be honest it's all under the hood, so I'm not really sure.

So in order for these ECS tasks to start talking to Snowflake again, we need to open up the security group to talk over more ports. Here I'm just allowing outgoing connections to the internet over any port (incoming requests are still restricted).

ian-r-rose (commented):
Okay, at long last, I think I'm unblocked here. I believe this is read for review!

melanie-logan (approved):
Reviewed and approved.

ian-r-rose:
Thanks for the review @melanie-logan!

PR #237, opened 2023-11-02
Autosuspend logging warehouse after one second

ian-r-rose:
Autosuspend logging warehouse even more aggressively to reduce costs as it logs every two minutes.


AeriShan-ODI (approved):
lgtm

PR #238, opened 2023-11-02
Bump terraform snowflake provider 

ian-r-rose:
Fix for an issue I just ran into when testing the native Snowflake GA4 connector: https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1962


AeriShan-ODI (approved):
👍

PR #241, opened 2023-11-16
Airflow maintenance and improvements

ian-r-rose:
In working on https://github.com/cagov/caldata-mdsa-caltrans-pems/issues/18, I found there were a few maintenance-related things I had to do to make our Airflow instance healthy after being unused for a few months:

* Add a convenience script for interacting with the MWAA CLI (for things that can't be done via the web UI)
* Turn on logging for better debugging when things go wrong
* Bump the version to 2.7.2
* Include an appropriate [constraints](https://airflow.apache.org/docs/apache-airflow/stable/installation/installing-from-pypi.html#constraints-files) file when specifying python dependencies.

### DIFF
--- a/airflow/aws_cli.py
+++ b/airflow/aws_cli.py
@@ -1,0 +1,34 @@
+"""
+Python script to run airflow CLI commands in a MWAA environment.
+
+Adapted from sample code here:
+https://docs.aws.amazon.com/mwaa/latest/userguide/airflow-cli-command-reference.html#airflow-cli-command-examples
+"""
+import base64
+import sys
+
+import boto3
+import requests
+
+mwaa_env_name = "dse-infra-dev-us-west-2-mwaa-environment"
+
+client = boto3.client("mwaa")
+
+mwaa_cli_token = client.create_cli_token(Name=mwaa_env_name)
+
+mwaa_auth_token = "Bearer " + mwaa_cli_token["CliToken"]
+mwaa_webserver_hostname = f"https://{mwaa_cli_token['WebServerHostname']}/aws_mwaa/cli"
+raw_data = " ".join(sys.argv[1:])
+
+mwaa_response = requests.post(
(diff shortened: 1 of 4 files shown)


airflow/aws_cli.py
ian-r-rose, line 1: This script isn't needed for daily operations, but I found it useful for debugging and doing various things with the Airflow CLI that aren't exposed through the web UI. Seemed useful to include it here.

airflow/deploy.sh
```diff
@@ -4,7 +4,12 @@ BUCKET="dse-infra-dev-us-west-2-mwaa"
 ENVIRONMENT="dse-infra-dev-us-west-2-mwaa-environment"
 
 # Copy the dags
-aws s3 sync dags s3://${BUCKET}/dags --delete --exclude "*__pycache__*"
+for SUBDIR in $(ls dags); do
```
ian-r-rose, line 7: This is a little less aggressive about `--delete`ing things in the `dags` directory: now we only delete things that belong to this repo's subdirectories.

This change is because I'm including some DAGs from other repos in the Airflow instance. It may not be considered best practice, or what we would have designed from scratch, but was expedient.

Perhaps worth having a discussion around multi-repo orchestrators?

airflow/requirements/requirements.txt
```diff
@@ -1,5 +1,9 @@
+--constraint https://raw.githubusercontent.com/apache/airflow/constraints-2.7.2/constraints-3.10.txt
 apache-airflow-providers-slack
-apache-airflow-providers-snowflake[slack]
```
ian-r-rose, line 2: Not sure what I was thinking here

AeriShan-ODI (approved):
A lot here and I reviewed everything but didn't have an opportunity to test. But no red or yellow flags...

ian-r-rose:
Thanks for the review!

PR #243, opened 2023-11-22
Add ruff formatter.

### LINKED ISSUE #239: Use `ruff` for formatting
Recent versions of `ruff` now support `black`-compatible formatting: https://docs.astral.sh/ruff/formatter/

Using `ruff` for this would simplify our pre-commit and CI pipelines

ian-r-rose:
This removes the black formatter in favor of ruff's allowing us to drop one tool . It's supposed to be a drop-in replacement, so no further updates are needed.

Fixes #239

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -42,18 +42,12 @@ repos:
             "--output-file",
             "../../../../docs/snowflake.md",
           ]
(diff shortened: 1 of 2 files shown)


AeriShan-ODI (approved):
LGTM

PR #249, opened 2023-12-05
Link to Fennis' dataset of errors-of-inclusion

ian-r-rose:
@fennisreed, I've added a few sentences on your errors-of-inclusion dataset and linked out to it. Is there anything else you'd want to include here?


fennisreed (approved):
We may want to add reference to the ArcToolbox here as well, as I've added a 'bad footprint' removal toggle to it. Otherwise it looks great to me! Thank you, Ian!

ian-r-rose:
Going to deploy this, thanks for the review and the updated toolbox files @fennisreed!

Issue #247, opened 2023-12-04
Improve Batch CI/CD


ian-r-rose:
We currently use AWS Batch for running jobs in docker containers. This allows us to run custom code with custom instance types, allowing for more flexibility and better compute scalability than running things in, e.g. Airflow worker processes. For instance, in an Airflow worker it can be very difficult to deal with python package constraints for things like pyarrow, geopandas, and snowflake-connector-python.

However, there are a few things that are difficult to work with in our current Batch set up:

1. It's not easy to build and push an image to a testing tag (so we don't overwrite `latest`)
2. It's not simple to trigger a job with custom container overrides
3. The test cycle is quite slow (build an image, pull it to an ecs container, run, wait for logs)

We should make some DX improvements! @melanie-logan might be interested in collaborating on some of this.

PR #257, opened 2023-12-28
Move building footprints pipeline to PRD snowflake environment

### LINKED ISSUE #251: Move building footprints DAG to PRD
As of now, most of our building footprints pipeline is targeting the `DEV` environment. We should move it to `PRD` as it is more-or-less finalized.

ian-r-rose:
Fixes #251 

This is a little more involved than the issue description would suggest: we have been running the data ingest in AWS Batch jobs, and our default job definition currently has access to our Snowflake LOADER_DEV role. So I've restructured our Batch set-up a little bit:

1. There are now *two* batch job definitions, a "test" definition, and a "latest". The "latest" definition has access to the PRD role, and the "test" definition has access to the DEV role.
2. Two different docker images are built with different tags: `latest` is built on merges to `main`, while `test` is built on pushes to feature branches. This is a small step towards implementing #247, though I wouldn't consider this fixing it.
3. The "test" job definition and docker image are intended to be used for developing new Batch jobs, "latest" is intended to be run in production by MWAA.

This new Batch set-up could have been done instead in a new AWS account, where we have stronger separation between dev/prod and a whole new Airflow deployment (cf. #119). That felt premature to me, however, and I think there's value in having some dev/test separation for batch jobs even in our "dev" environment. I'm open to push-back here, though!

### DIFF
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -1,9 +1,6 @@
-name: deploy
+name: Build Image
 
-on:
-  push:
-    branches:
-      - main
+on: push
 
 jobs:
   deploy:
@@ -33,7 +30,7 @@ jobs:
         env:
           REGISTRY: ${{ steps.login-ecr.outputs.registry }}
           REPOSITORY: dse-infra-dev-us-west-2-default
-          IMAGE_TAG: latest
+          IMAGE_TAG: ${{ github.ref == 'refs/heads/main' && 'latest' || 'test' }}
         with:
           push: true
           context: "."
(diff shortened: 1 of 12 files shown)


ian-r-rose:
Okay, I think this is ready for a look!

melanie-logan (approved):
I agree, I think there's value in having some dev/test separation for batch jobs in our "dev" environment.

The config updates look good to me.

britt-allen (approved):
This is all makes sense and looks good to me. TIL MWAA. Thank you Ian!

ian-r-rose:
Thanks for the reviews!

PR #266, opened 2024-01-05
Schedule downstream footprint tasks

### LINKED ISSUE #250: Incorporate downstream tasks in building footprints pipeline
With #233 more-or-less complete, we should be able add a couple of downstream tasks to the pipeline, making it more "production" grade.

- After the loading step is done, we can trigger the dbt build of the footprints-with-tiger-data. We could either use dbt core for this, or a dbt cloud operator.
- After the dbt build is done, we can trigger the unload-to-s3 step.

ian-r-rose:
Fixes #250 

Previously, we were only scheduling the data load of building footprints. The dbt build was left for a "nightly" job, and the data unload to geoparquet/shapefile was done manually.

This incorporates the dbt build for footprints as well as the data unloading step into the Airflow dag so that it's more of an end-to-end pipeline. I'm triggering the dbt build using the `DbtCloudJobRunOperator`, which talks directly to environments and jobs in dbt cloud. We could also do this with dbt core, and reworking things to do that would be an interesting exercise. For the moment, however, using dbt cloud was simpler.

### DIFF
--- a/.github/workflows/submit-job.yml
+++ b/.github/workflows/submit-job.yml
@@ -25,10 +25,8 @@ jobs:
           aws batch submit-job \
           --job-name test \
           --job-queue dse-infra-dev-us-west-2-default \
-          --job-definition dse-infra-dev-us-west-2-default-latest \
+          --job-definition dse-infra-dev-us-west-2-latest \
           --container-overrides '{
             "resourceRequirements":
-            [{"value": "2", "type": "VCPU"}, {"value": "4096", "type": "MEMORY"}],
-            "environment":
-            [{"name": "BUCKET", "value": "dse-infra-dev-us-west-2-scratch"}]
+            [{"value": "2", "type": "VCPU"}, {"value": "4096", "type": "MEMORY"}]
           }'
--- a/airflow/dags/geo_reference/load_building_footprints.py
+++ b/airflow/dags/geo_reference/load_building_footprints.py
@@ -63,7 +63,7 @@ def building_footprints_dag():
     )
 
     run_dbt_cloud_job = DbtCloudRunJobOperator(
-        job_id=None,
+        job_id=14,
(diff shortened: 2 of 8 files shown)


.github/workflows/submit-job.yml
ian-r-rose, line 1: This is a drive-by fix of this github action, which was inadvertently broken by #257. The action here is more of a demonstration/smoke-test of submitting batch jobs via github actions, and could be deleted.

airflow/dags/geo_reference/load_building_footprints.py
```diff
@@ -9,6 +9,24 @@
 from airflow.decorators import dag
 from airflow.providers.amazon.aws.operators.batch import BatchOperator
 from airflow.providers.amazon.aws.sensors.batch import BatchSensor
+from airflow.providers.dbt.cloud.operators.dbt import DbtCloudRunJobOperator
+
+
+def _construct_batch_args(name: str, command: list[str]) -> dict:
```
ian-r-rose, line 15: Small refactor to try to avoid repeating the same batch job args over and over

terraform/aws/modules/infra/batch.tf
```diff
@@ -84,6 +84,11 @@ resource "aws_iam_role_policy_attachment" "s3_scratch_policy_role_attachment" {
   policy_arn = aws_iam_policy.s3_scratch_policy.arn
 }
 
+resource "aws_iam_role_policy_attachment" "dof_demographics_read_write_access" {
```
ian-r-rose, line 87: This gives our Batch ECS tasks write access to the DOF demographics bucket

melanie-logan (approved):
Updates look good Ian!

ian-r-rose:
Thanks for the review @melanie-logan!

PR #278, opened 2024-02-13
changed/added configuration for yaml linting

### LINKED ISSUE #279: Enable proseWrap for prettier
[This](https://prettier.io/docs/en/options.html#prose-wrap) will make authoring long prose strings in YAML easier.

We should also bump the line length to 120 and set `endOfLine: auto` while we are at it.

Related PR #278

britt-allen:
Closes #279

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -69,6 +69,8 @@ repos:
     hooks:
       - id: prettier
         types: [yaml]
+        endOfLine: auto
+        proseWrap: "always"
   - repo: https://github.com/adrienverge/yamllint.git
     rev: v1.28.0
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -69,8 +69,6 @@ repos:
     hooks:
       - id: prettier
         types: [yaml]
-        endOfLine: auto
-        proseWrap: "always"
   - repo: https://github.com/adrienverge/yamllint.git
     rev: v1.28.0
     hooks:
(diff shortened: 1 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -69,6 +69,8 @@ repos:
     hooks:
       - id: prettier
         types: [yaml]
+        endOfLine: auto
+        proseWrap: "always"
```
ian-r-rose, line 73: Prettier config goes into a `.prettierrc` file. For an example, see the CalHR repo

britt-allen: Thank you, will take a look. The docs and instructions were fairly vague on where this should go.

ian-r-rose: Here is documentation on how to configure prettier: https://prettier.io/docs/en/configuration

britt-allen: Thanks for the link, I looked at that earlier. It wasn't clear to me that I should be creating a new configuration file to add these to versus adding these arguments to existing files. I searched our repo for `# .prettierrc or .prettierrc.yaml` and couldn't find it so I thought we were not in need of the file. Good to know I should have just created it. Thanks again!

ian-r-rose:
Looks like it wants to shorten a bunch of strings in our YAML. Which is good! But it looks like it's probably wrapping at 80 characters. For consistency, does it have a lighter hand if we also make [`printWidth`](https://prettier.io/docs/en/options.html#print-width) 120 characters?

.pre-commit-config.yaml
ian-r-rose: Yeah, we need to create it. It would be technically possible to pass it in via CLI args in the pre-commit config, but I think that's not the best way to do it

britt-allen:
> Looks like it wants to shorten a bunch of strings in our YAML. Which is good! But it looks like it's probably wrapping at 80 characters. For consistency, does it have a lighter hand if we also make [`printWidth`](https://prettier.io/docs/en/options.html#print-width) 120 characters?

We (Ian and I - chatted offline) decided this behavior is not what we want. We will preserve user-input rather than always auto-wrapping at a specific print width. This was causing a lot of old files to fail pre-commit checks as well.

ian-r-rose (approved):
LGTM, thanks!

PR #318, opened 2024-05-20
Private key variable

JamesSLogan:
Let me know if there are other areas that need to be updated. A quick search on the word "private" shows some docs that don't appear to need to be changed since they speak to Terraform setup.

I also increased the cache action version since it was working in user-journey-tracking.

----
Note for posterity:
I ran into an issue connecting to Snowflake when upgrading the user-journey-tracking repo to use this code. The error I was seeing was "Incorrect padding", and it appears to be [this issue](https://github.com/dbt-labs/dbt-snowflake/issues/640). What worked for me was to re-add/rotate keys as such:
1. Add private key as github secret _with_ line breaks but no `-----BEGIN/END PRIVATE KEY-----` 
2. Add public key to snowflake _without_ line breaks and also no `-----BEGIN/END PRIVATE KEY-----` (I think this works even with line breaks...)


ian-r-rose:
You may need to bump the version of `dbt-core` and `dbt-snowflake` in the `pyproject.toml` for this to work (I think my note about version 1.5 wound up not being correct, it was more like 1.6.x).

> Add public key to snowflake without line breaks and also no -----BEGIN/END PRIVATE KEY----- (I think this works even with line breaks...)

Snowflake *used* to require that you manually remove line breaks, but thankfully some service update during the last year allowed them. Much easier to set a public key now.

JamesSLogan:
> You may need to bump the version of dbt-core and dbt-snowflake in the pyproject.toml for this to work (I think my note about version 1.5 wound up not being correct, it was more like 1.6.x).

The current versions are sufficient 👍. I did make an attempt to update these packages to more recent versions but ran into a fair number of landmines with sqlfluff.

ian-r-rose (approved):
Thanks @JamesSLogan

PR #351, opened 2024-08-21
docs build update to address #267

### LINKED ISSUE #267: Build docs on PRs
Over in https://github.com/cagov/caldata-mdsa-carb-enforcement/pull/177 we started building the docs on PRs as well as on `main` (we still only deploy them to github pages on `main`). This helps to catch any docs build issues before they land. It would be nice to port that logic over here!

britt-allen:
fixes #267

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,8 +1,9 @@
 name: docs
 on:
+  pull_request:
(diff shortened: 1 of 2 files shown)


JamesSLogan (approved):
Looks great

PR #389, opened 2024-10-01
Reorganized documentation

### LINKED ISSUE #347: Improve information architecture of our docs
Does the current structure of pages make sense right now? Probably not, it accreted over time, without much thought to what a sensible organization might be. A lot of the names of subpages are confusing, and relevant information on similar topics is scattered across a number of different pages. For instance, information about setting up Snowflake can be found [here](https://cagov.github.io/data-infrastructure/setup/#configure-snowflake), [here](https://cagov.github.io/data-infrastructure/snowflake/), and [here](https://cagov.github.io/data-infrastructure/dbt/).

Let's think about how we can reorganize the pages and the content therein to make it easier to navigate.

JamesSLogan:
This PR would establish a more hierarchical structure for documentation using a few different categories. No documentation has been added or removed, just shifted around. 

Fixes #347


britt-allen (approved):
This is incredibly helpful and in line with some changes we were just discussing as a team. Thank you so much James!

britt-allen:
I think you have some white space issues in mkdocs.yml @JamesSLogan

ian-r-rose:
Thank you @JamesSLogan, the docs look _much_ more legible now!

PR #398, opened 2024-10-10
Update terraform CI to execute in its own job

### LINKED ISSUE #319: Don't require terraform set up to run pre-commit checks

JamesSLogan:
Fixes #319 

This ended up being more painful than the related task in Caltrans' repo due to the terraform-docs automation. The terraform-docs action works for our use case, though it requires 1 invocation per file and also has a bug [(see use of $OUTPUT_FILE here)](https://github.com/terraform-docs/gh-actions/pull/108) that means it only fully works on 1 of our 3 files.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -23,14 +23,6 @@ jobs:
       - uses: snok/install-poetry@v1
         with:
           virtualenvs-create: false
-      - name: Setup terraform
-        uses: hashicorp/setup-terraform@v2
-        with:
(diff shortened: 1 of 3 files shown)


ian-r-rose:
I'm also open to deleting the terraform-docs functionality, I haven't found it that useful. But if you figured it out, then that's great too!

JamesSLogan:
I guess I would leave that up to you all! It should work either way...I did test the scenarios where files do get changed.

britt-allen (approved):

PR #407, opened 2024-10-16
doc updates

britt-allen:


docs/code/local-setup.md
```diff
@@ -11,8 +11,9 @@ Much of the software in this project is written in Python.
 It is usually worthwhile to install Python packages into a virtual environment,
 which allows them to be isolated from those in other projects which might have different version constraints.
 
-One popular solution for managing Python environments is [Anaconda/Miniconda](https://docs.conda.io/en/latest/miniconda.html).
-Another option is to use [`pyenv`](https://github.com/pyenv/pyenv).
+Most of our team uses [Anaconda](https://docs.anaconda.com/anaconda/install/).
+Another popular solution for managing Python environments is [Anaconda/Miniconda](https://docs.conda.io/en/latest/miniconda.html).
```
ian-r-rose, line 15: This is confusing to me, since Miniconda is a variation of Anaconda, and you list Anaconda twice.

docs/code/local-setup.md
```diff
@@ -33,7 +34,7 @@ Here are instructions for setting up a Python environment using Miniconda:
 
 Python dependencies are specified using [`poetry`](https://python-poetry.org/).
 
-To install them, open a terminal and ensure you are working in the data-infrastructure root folder, then enter the following:
+To install them, open a terminal and ensure you are working in the (`data-infrastructure`) root folder with your `infra` environment activated, then enter the following:
```
ian-r-rose, line 37: Why wrap this in parentheses?

docs/code/local-setup.md
```diff
@@ -52,6 +53,8 @@ First, install Homebrew
 
 ```bash
 /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+
+export PATH=/opt/homebrew/bin:/opt/homebrew/sbin:$PATH
```
ian-r-rose, line 57: I don't think this does what you want, as it only temporarily updates the path, and may miss other things that homebrew wants to be initialized in your shell. The install script gives instructions for how to add the activation to your bash or zsh profile, did they not work for you?

docs/code/local-setup.md
```diff
@@ -117,29 +118,32 @@ you need to create access keys and configure your local setup to use them:
...
-so should not be checked into version control.
+This is because connection information is both user-specific and usually sensitive,
+so it should not be checked into version control.
+
 In order to run this project locally, you will need to provide this information
-in a YAML file located (by default) in `~/.dbt/profiles.yml`.
+in a YAML file. Run the following command to create the necessary folder and file.
+
+```bash
+mkdir ~/.dbt && touch ~/.dbt/profiles.yml
```
ian-r-rose, line 130: 👍 Nice addition. Maybe we should call out that this is only expected to work on posix-y systems, and Windows users will have a different command. In general, I have found this to be a very challenging step for some users who are unaccustomed to creating folders and files in specific places.

ian-r-rose (commented):
Nice updates!

docs/code/local-setup.md
britt-allen: I think I missed those instructions personally.

docs/code/local-setup.md
britt-allen: Thank you for that note!

britt-allen:
Yup, planning to pull them in once this merges

ian-r-rose (approved):

PR #413, opened 2024-10-24
Changes for upgrading terraform snowflake provider from 0.92 to 0.97

ram-kishore-odi:
After the initial setup and validation, the main change needed was to change snowflake_role to snowflake_account_role. 
Here is the partial output from terraform plan, Please review it and share your feedback.

****************************************************

  # module.elt.module.transforming["XL"].snowflake_warehouse.this will be updated in-place
  ~ resource "snowflake_warehouse" "this" {
      ~ enable_query_acceleration           = "false" -> "default"
        id                                  = "TRANSFORMING_XL_DEV"
      - max_cluster_count                   = 1 -> null
      ~ max_concurrency_level               = 8 -> (known after apply)
      - min_cluster_count                   = 1 -> null
        name                                = "TRANSFORMING_XL_DEV"
      ~ query_acceleration_max_scale_factor = 8 -> -1
      - scaling_policy                      = "STANDARD" -> null
      ~ show_output                         = [
          - {
              - auto_resume                         = true
              - auto_suspend                        = 300
              - available                           = 0
              - comment                             = "Primary warehouse for transforming data. Analytics engineers and automated transformation tools should use this warehouse"
              - created_on                          = "2023-05-25 15:24:30.262 -0700 PDT"
              - enable_query_acceleration           = false
              - is_current                          = false
              - is_default                          = false
              - max_cluster_count                   = 1
              - min_cluster_count                   = 1
              - name                                = "TRANSFORMING_XL_DEV"
              - other                               = 0
              - owner                               = "SYSADMIN"
              - owner_role_type                     = "ROLE"
              - provisioning                        = 0
              - query_acceleration_max_scale_factor = 8
              - queued                              = 0
              - quiescing                           = 0
              - resumed_on                          = "2024-05-02 12:02:40.752 -0700 PDT"
              - running                             = 0
              - scaling_policy                      = "STANDARD"
              - size                                = "XLARGE"
              - started_clusters                    = 0
              - state                               = "SUSPENDED"
              - type                                = "STANDARD"
              - updated_on                          = "2024-05-02 12:02:40.752 -0700 PDT"
                # (1 unchanged attribute hidden)
            },
        ] -> (known after apply)
      ~ statement_timeout_in_seconds        = 172800 -> (known after apply)
      - warehouse_type                      = "STANDARD" -> null
        # (8 unchanged attributes hidden)
    }

  # module.elt.module.transforming["XS"].snowflake_account_role.this will be created
  + resource "snowflake_account_role" "this" {
[...]

PR #418, opened 2024-11-06
Rework warehouse `for_each` statements

ian-r-rose:
Follow-up to #413 

The existing `for_each` construction wound up causing problems because terraform was unable to statically determine the keys for the for loop (I guess the expression was just a bit too complex, some more discussion [here](https://discuss.hashicorp.com/t/invalid-for-each-argument-during-import/55332/6)).

This PR simplifies the `for_each` statements a bit which seems to resolve the problem. The actual content of the configuration does not change, though the grants need to be recreated since they have different names (this has less impact than recreating roles).


ram-kishore-odi (approved):
Changes look good to me !

PR #420, opened 2024-11-13
Added additional warehouse size options

ram-kishore-odi:
Can you please review the changes and let me know your feedback ?


britt-allen (approved):
Looks great, thank you Ram!

ian-r-rose (approved):
Looks good, thanks @ram-kishore-odi

PR #435, opened 2024-11-25
dbt versionless exploration

summer-mothwood:
testing some changes for dbt versionless:

* changed syntax for tests (not required, but a proactive measure!)
* defined the following new behavior flags:

`  state_modified_compare_more_unrendered_values: true` - adding this flag and setting it to true, then running a full refresh, will hopefully improve CI when we do this in the Caltrans project

  `skip_nodes_if_on_run_start_fails: true` - unsure if we want this flag as true or false. if true, it will skip running anything if any of the on_run_start hooks fails. 
  
  `require_explicit_package_overrides_for_builtin_materializations: true` - this seems like a no-brainer extra security layer that we might as well turn on sooner rather than later. more details here if interested: https://github.com/dbt-labs/dbt-core/security/advisories/GHSA-p3f3-5ccg-83xq
  
  `source_freshness_run_project_hooks: false` - not sure if we want to turn this on or not. I assume we do not want to / need to check source freshness for our project hooks (in here or in Caltrans), but lmk! https://docs.getdbt.com/reference/global-configs/behavior-changes#project-hooks-with-source-freshness

I left all other new behavior flags out as they either don't apply to our project, or they did not seem important enough to test right now:
* https://docs.getdbt.com/reference/global-configs/behavior-changes#no-spaces-in-resource-names - this flag shouldn't be triggered by any of our models, so we can probably just wait for this behavior to become enabled by default 
* https://docs.getdbt.com/reference/global-configs/behavior-changes#metricflow-time-spine-yaml - we do not use Metric Flow for our time spines in the Caltrans project as far as I can tell
* https://docs.getdbt.com/reference/global-configs/behavior-changes#custom-microbatch-strategy - we do not have any custom-built microbatch strategies, so no need to bother with this flag. we'll be able to test microbatching without it, this flag is only for people who were already playing around with building their own version of microbatching.


ian-r-rose (approved):
LGTM, thanks @summer-mothwood

ian-r-rose:
We may also want to bump the dbt version here: https://github.com/cagov/data-infrastructure/blob/2bdfdbb98b7c0aef8011017546e9099f0074468e/pyproject.toml#L67-L68

I'm not sure what keeping these versions in sync looks like in a "versionless" world (though we weren't exactly doing a great job of keeping them in sync in a versioned world)

summer-mothwood:
> We may also want to bump the dbt version here:
> 
> https://github.com/cagov/data-infrastructure/blob/2bdfdbb98b7c0aef8011017546e9099f0074468e/pyproject.toml#L67-L68
> 
> I'm not sure what keeping these versions in sync looks like in a "versionless" world (though we weren't exactly doing a great job of keeping them in sync in a versioned world)

Yes this is a good point! I'm not sure how we should handle this in poetry actually. Because in dbt Core version 1.8+, adapters are de-coupled from dbt-core, we should probably make the changes to include dbt-adapters and dbt-common as outlined here https://github.com/dbt-labs/dbt-adapters/discussions/87

In that doc they recommend specifying dbt-core version as "1.8.0+", but specified in a comment later in the thread that in the future, we won't need dbt-core specified there at all (at least for the adapter dependencies). 

What do you think? I'm new to setting up poetry dependencies so I might be missing something!

ian-r-rose:
I actually thought that 1.9 had been already released, but I now see that it hasn't so perhaps we don't have to touch poetry right now (we could still upgrade within the 1.8 series if we wanted to).

> Because in dbt Core version 1.8+, adapters are de-coupled from dbt-core, we should probably make the changes to include dbt-adapters and dbt-common as outlined here https://github.com/dbt-labs/dbt-adapters/discussions/87

Based on my read of that discussion, `dbt-adapters` and `dbt-common` are both transitive dependencies for `dbt-core` and `dbt-snowflake`, so I'm not sure any action would be required to ensure that those are brought in.

> In that doc they recommend specifying dbt-core version as ">=1.8.0"

I disagree with this recommendation. Setting `>=1.8.0` means that we could accidentally upgrade to 1.9 or 2.0, and I'm not sure we trust the stability of `dbt-core` enough for that right now. Using the `~` specifier keeps us in the 1.8 series.

> but specified in a comment later in the thread that in the future, we won't need dbt-core specified there at all (at least for the adapter dependencies).

I don't think this is true: after they decoupled the adapter from core, it makes it more important to bring in `dbt-core` explicitly, since that's what carries the main logic and CLI.

summer-mothwood:
> I actually thought that 1.9 had been already released, but I now see that it hasn't so perhaps we don't have to touch poetry right now (we could still upgrade within the 1.8 series if we wanted to).
> 
> > Because in dbt Core version 1.8+, adapters are de-coupled from dbt-core, we should probably make the changes to include dbt-adapters and dbt-common as outlined here [dbt-labs/dbt-adapters#87](https://github.com/dbt-labs/dbt-adapters/discussions/87)
> 
> Based on my read of that discussion, `dbt-adapters` and `dbt-common` are both transitive dependencies for `dbt-core` and `dbt-snowflake`, so I'm not sure any action would be required to ensure that those are brought in.
> 
> > In that doc they recommend specifying dbt-core version as ">=1.8.0"
> 
> I disagree with this recommendation. Setting `>=1.8.0` means that we could accidentally upgrade to 1.9 or 2.0, and I'm not sure we trust the stability of `dbt-core` enough for that right now. Using the `~` specifier keeps us in the 1.8 series.
> 
> > but specified in a comment later in the thread that in the future, we won't need dbt-core specified there at all (at least for the adapter dependencies).
> 
> I don't think this is true: after they decoupled the adapter from core, it makes it more important to bring in `dbt-core` explicitly, since that's what carries the main logic and CLI.

Ah yeah, I thought we were in 1.9 when I was deep in the docs last week, because a lot of Versionless documentation has lines like 'this is available in Versionless and 1.9'...sort of confusing.

I went back to the docs to look for more info, and found this line I missed before "[Note that adapters running in dbt Cloud **must** be migrated to the new decoupled architecture as a baseline in order to support the new "Versionless" option.](https://docs.getdbt.com/blog/latest-dbt-stability#decoupling-the-adapter-interface)"

ian-r-rose:
I think that given 1.8 is still the most recent version, we can move forward as is, thanks @summer-mothwood!

PR #448, opened 2024-12-12
Update URLs to their new home

### LINKED ISSUE #434: Update links to building footprints source dataset
The URL for the GlobalML building footprints [has changed](https://github.com/microsoft/GlobalMLBuildingFootprints?tab=readme-ov-file#updates). We'll need to update the links to it in our data ingest script.

ian-r-rose:
Fixes #434 

Interestingly, the US Building Footprints URLs now have "legacy" in them, a seeming confirmation of what we suspected about the future of that dataset.


ram-kishore-odi (approved):
The changes look good to me.

PR #485, opened 2025-03-10
Inital default password policy changes

ram-kishore-odi:
Can you please review the changes and share your feedback ?

The changes include 

1. Defining a default password policy
2. Attaching it to the account

Also I could not convert the grant below into terraform script. I do not think this a blocker and it is not critical.
May be I am missing something. I thought it should be simple to do this. But in the current documentation, I did not see an option available to do this in tf.

 -- GRANT USAGE ON PASSWORD POLICY user_password_policy TO ROLE securityadmin; ---

Looking at the plan looks like the service account are getting impacted, please refer to the attached plan.
I am concerned about the changes to service account password that would be needed.

### DIFF
--- a/terraform/snowflake/environments/dev/main.tf
+++ b/terraform/snowflake/environments/dev/main.tf
@@ -42,6 +42,7 @@ provider "snowflake" {
   account_name      = var.account_name
   organization_name = var.organization_name
   role    = "PUBLIC"
+  preview_features_enabled = ["snowflake_authentication_policy_resource", "snowflake_password_policy_resource"]
 }
 
 # Snowflake provider for account administration (to be used only when necessary).
@@ -50,6 +51,7 @@ provider "snowflake" {
   role    = "ACCOUNTADMIN"
   account_name      = var.account_name
   organization_name = var.organization_name
+  preview_features_enabled = ["snowflake_authentication_policy_resource", "snowflake_password_policy_resource"]
 }
 
 # Snowflake provider for creating databases, warehouses, etc.
@@ -58,6 +60,7 @@ provider "snowflake" {
   role    = "SYSADMIN"
   account_name      = var.account_name
   organization_name = var.organization_name
+  preview_features_enabled = ["snowflake_authentication_policy_resource", "snowflake_password_policy_resource"]
 }
 
 # Snowflake provider for managing grants to roles.
@@ -66,6 +69,7 @@ provider "snowflake" {
   role    = "SECURITYADMIN"
   account_name      = var.account_name
   organization_name = var.organization_name
+  preview_features_enabled = ["snowflake_authentication_policy_resource", "snowflake_password_policy_resource"]
 }
 
 # Snowflake provider for managing user accounts and roles.
@@ -74,6 +78,7 @@ provider "snowflake" {
   role    = "USERADMIN"
   account_name      = var.account_name
   organization_name = var.organization_name
+  preview_features_enabled = ["snowflake_authentication_policy_resource", "snowflake_password_policy_resource"]
 }
 
 ############################
@@ -103,4 +108,26 @@ resource "snowflake_grant_account_role" "logger_to_transformer" {
   provider         = snowflake.useradmin
   role_name        = "LOGGER_${var.environment}"
   parent_role_name = "TRANSFORMER_${var.environment}"
+}
+
+# Default user password policy
+resource "snowflake_password_policy" "user_password_policy" {
+  database             = "POLICIES" # Database name
+  schema               = "PUBLIC"   # Schema name
+  name                 = "user_password_policy"
+  min_length           = 14
+  min_upper_case_chars = 1
+  min_lower_case_chars = 1
+  min_numeric_chars    = 1
+  min_special_chars    = 1
+  max_retries          = 5
+  lockout_time_mins    = 30
+  history              = 5
+  max_age_days         = 60
+  or_replace           = true # Ensures the policy can be updated without errors
+}
+
+# Set the default password policy for the account
+resource "snowflake_account_password_policy_attachment" "attachment" {
+  password_policy = snowflake_password_policy.user_password_policy.fully_qualified_name
 }


terraform/snowflake/environments/dev/main.tf
```diff
@@ -104,3 +109,25 @@ resource "snowflake_grant_account_role" "logger_to_transformer" {
   role_name        = "LOGGER_${var.environment}"
   parent_role_name = "TRANSFORMER_${var.environment}"
 }
+
+# Default user password policy
+resource "snowflake_password_policy" "user_password_policy" {
```
ian-r-rose, line 114: This probably defaults to the "PUBLIC" provider above, which won't have permissions to set this policy.
It might need something like the following:
```suggestion
resource "snowflake_password_policy" "user_password_policy" {
  provider = snowflake.securityadmin
```

terraform/snowflake/environments/dev/main.tf
```diff
@@ -104,3 +109,25 @@ resource "snowflake_grant_account_role" "logger_to_transformer" {
   role_name        = "LOGGER_${var.environment}"
   parent_role_name = "TRANSFORMER_${var.environment}"
 }
+
+# Default user password policy
```
ian-r-rose, line 113: For account-wide things like this, I think it makes more sense to put it in the `prd` environment (long term, it may be a good idea to make a new module for these, actually)

terraform/snowflake/environments/dev/main.tf
```diff
@@ -42,6 +42,7 @@ provider "snowflake" {
   account_name      = var.account_name
   organization_name = var.organization_name
   role    = "PUBLIC"
+  preview_features_enabled = ["snowflake_authentication_policy_resource", "snowflake_password_policy_resource"]
```
ian-r-rose, line 45: Not a huge deal, but I think only `securityadmin` or `accountadmin` would need these preview features. Do you intend to include authentication policies here as well?

ian-r-rose (commented):
Looking forward to having this in!

ian-r-rose:
> Also I could not convert the grant below into terraform script. I do not think this a blocker and it is not critical. May be I am missing something. I thought it should be simple to do this. But in the current documentation, I did not see an option available to do this in tf.
> 
> -- GRANT USAGE ON PASSWORD POLICY user_password_policy TO ROLE securityadmin; ---

Does a grant similar to this one not work? https://github.com/cagov/data-infrastructure/blob/4cb9f93e032912bd9062dc36d48ad78a26278fb2/terraform/snowflake/modules/elt/roles.tf#L187-L195

ian-r-rose:
> Looking at the plan looks like the service account are getting impacted, please refer to the attached plan.

I don't see an attached plan. Weird changes to service account metadata is also something I have seen. I wish I knew what was causing it! But it hasn't caused problems for me in the past when I apply them

terraform/snowflake/environments/dev/main.tf
ram-kishore-odi: Thank you Ian ! Good catch !

terraform/snowflake/environments/dev/main.tf
ram-kishore-odi: I agree, as we discussed yesterday, I will move this to prd instead of dev. Also I plan to create a new story for the point you raised about separating this functionality / moving out of main

terraform/snowflake/environments/dev/main.tf
```diff
@@ -104,3 +109,26 @@ resource "snowflake_grant_account_role" "logger_to_transformer" {
   role_name        = "LOGGER_${var.environment}"
   parent_role_name = "TRANSFORMER_${var.environment}"
 }
+
+# Default user password policy
+resource "snowflake_password_policy" "user_password_policy" {
+  provider = snowflake.securityadmin
+  database             = "POLICIES" # Database name
```
ian-r-rose, line 116: Does something create this `POLICIES` database in the configuration? I don't think it does. If so, we need to either incorporate a `policies` database into this as a resource or as a data source. Otherwise it would be difficult to stand-up a new environment in a new account whenever we need to do that next

terraform/snowflake/environments/dev/main.tf
ram-kishore-odi: I agree, securityadmin or accountadmin would need these. Yes it is already WIP and I plan to add the authentication policies here (i.e. after I move this prd)

ian-r-rose: Thanks! Really looking forward to having both here

ram-kishore-odi:
> > Also I could not convert the grant below into terraform script. I do not think this a blocker and it is not critical. May be I am missing something. I thought it should be simple to do this. But in the current documentation, I did not see an option available to do this in tf.
> > -- GRANT USAGE ON PASSWORD POLICY user_password_policy TO ROLE securityadmin; ---
> 
> Does a grant similar to this one not work?
> 
> https://github.com/cagov/data-infrastructure/blob/4cb9f93e032912bd9062dc36d48ad78a26278fb2/terraform/snowflake/modules/elt/roles.tf#L187-L195



> resource "snowflake_grant_privileges_to_account_role" "imported_privileges_to_logger" { 
>    provider          = snowflake.accountadmin 
>    account_role_name = snowflake_account_role.logger.name 
>    privileges        = ["IMPORTED PRIVILEGES"] 
>    on_account_object { 
>      object_type = "DATABASE" 
>      object_name = "SNOWFLAKE" 
>    } 
>  }

This is very helpful @ian-r-rose

terraform/snowflake/environments/dev/main.tf
ram-kishore-odi: Hi Ian,  POLICIES database and the PUBLIC schema are available by default. I saw those being used in example code snippets online as well as validated that they are available in CalHFA, ODI main and CalTrans instances. So I think we may not need to add a resource definition for it. Please let me know if you prefer an explicit definition to be added. If yes, it may be better to choose a different name

ram-kishore-odi:
Here is the latest plan - 
[Jsr-pp-prd.txt](https://github.com/user-attachments/files/19219244/Jsr-pp-prd.txt)

I think the grant role for usage on a password policy is not yet supported. Please refer to the error message

terraform/snowflake/environments/dev/main.tf
ian-r-rose: I remember now: `POLICIES` is *not* created by default, but is instead done manually as part of our account setup procedure. At the time we did this because the terraform provider didn't support session policies (it may now, however!).

https://cagov.github.io/data-infrastructure/setup/snowflake-setup/#apply-a-session-policy

I think our options here are either:

1. Move the policy database creation into the terraform configuration
2. Make sure we create the policies database before we apply our terraform configurations

.gitignore
```diff
@@ -161,3 +161,6 @@ cython_debug/
 #  and can be added to the global gitignore or merged into this file.  For a more nuclear
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
 #.idea/
+.DS_Store
+terraform/snowflake/environments/.DS_Store
+terraform/snowflake/environments/dev/.DS_Store
```
ian-r-rose, line 166: Doesn't the first `.DS_Store` already ignore subfolders?

terraform/snowflake/environments/prd/main.tf
```diff
@@ -104,3 +106,36 @@ resource "snowflake_grant_account_role" "logger_to_transformer" {
...
+# Granting USAGE to security admin on the password policy
+resource "snowflake_grant_privileges_to_account_role" "grant_usage_to_securityadmin" {
+  provider          = snowflake.accountadmin
+  account_role_name = "SECURITYADMIN" #  SECURITYADMIN role name
+  privileges        = ["USAGE"]
+  on_account_object {
+    object_type = "PASSWORD POLICY" # object type is a password policy
+    object_name = snowflake_password_policy.user_password_policy.name
+  }
+}
```
ian-r-rose, line 141: If this doesn't work yet, maybe we just have to skip the role grant and have it be owned by the role that creates it (either SECURITYADMIN or ACCOUNTADMIN). For a singleton policy, that doesn't seem so bad

.gitignore
ram-kishore-odi: Yes, you're correct. The top-level .DS_Store entry already ignores .DS_Store files in all subfolders. I'll remove terraform/snowflake/environments/dev/.DS_Store in the next commit

terraform/snowflake/environments/prd/main.tf
ram-kishore-odi: This sounds good ! I will comment out the grant.

ian-r-rose: Let's just remove it, no need to have dead code

PR #498, opened 2025-04-02
Update cache action to pick up poetry environment stored in .venv

ian-r-rose:
Working towards https://github.com/cagov/caldata-dse-internal-tracker/issues/68


sqlfluff_pre_commit.py
```diff
@@ -0,0 +1,33 @@
+# sqlfluff_pre_commit.py
```
summer-mothwood, line 1: This is interesting! Is this something we'd want to add to our other repos as well, or is there a reason we need the sqlfluff run w/ settings split out like this that is specific to this repo?

ian-r-rose: Yes, it's already in some of our other repos! It just seems like it hasn't made its way into our template yet. I just copied this file from one of the others, but will try to add it to the template as part of this effort

ian-r-rose: It may also be that this file is no longer necessary: it's a weird workaround because sqlfluff didn't work well with dbt projects in a repo subdirectory. I should test and see if more recent versions have fixed that

ian-r-rose: Thank for the nudge! I don't think this workaround script is necessary anymore, as it's now easier to invoke sqlfluff directly from the root. This PR is now simpler

ian-r-rose:
I think this is now ready for final revew

summer-mothwood (approved):
Lovely, thank you Ian!

Issue #521, opened 2025-06-04
Update Data Access and Grant Documentation


ram-kishore-odi:
**As a** Developer/Data Analyst,
**I need** updated documentation on database access and cross-environment grants
**So that** I can correctly develop and deploy Streamlit applications and manage data transformations effectively.

---

### Updates Needed:

* Outline the purpose and use cases for granting `RAW_PRD_READ` to `TRANSFORMER_DEV`.
* Detail the necessity and procedure for granting `TRANSFORM_PRD_READ` and `ANALYTICS_PRD_READ` to `TRANSFORMER_DEV`, especially for projects with large tables requiring cloning and deferral.
* Explain the granting of `ANALYTICS_PRD_READ` to `REPORTER_DEV` to support Streamlit development.
* Specify that all objects (marts, tables) related to production Streamlit dashboards must reside within the `ANALYTICS_ENGCA_PRD` database.
* Clarify that access to data in earlier layers for Streamlit dashboards should be provided via explicitly created mart tables.
* Include the manual SQL commands for the cross-environment grant of the `${module.raw.name}_${var.environment}_STREAMLIT` role to the `REPORTER_DEV` role and explain this process is managed outside of Terraform.

This is follow up story and related to this one - https://github.com/cagov/caldata-engaged-california/issues/26

PR #529, opened 2025-06-18
Updates to Data Access and Grant Documentation

ram-kishore-odi:
Can you please review the changes and let me know your feedback ?

The updates in this PR are related to this story - https://github.com/cagov/data-infrastructure/issues/521

### DIFF
--- a/docs/setup/snowflake-setup.md
+++ b/docs/setup/snowflake-setup.md
@@ -128,7 +128,7 @@ alter account set session policy policies.public.account_session_policy;
 
 ### Developing against production data
 
-Our Snowflake architecture allows for reasonably safe `SELECT`ing from the production `RAW` database while developing models.
+Our Snowflake architecture allows for reasonably safe `SELECT`ing from the production databases while developing models.
 While this could be expensive for large tables, it also allows for faster and more reliable model development.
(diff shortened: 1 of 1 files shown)


docs/setup/snowflake-setup.md
```diff
@@ -128,29 +128,42 @@ alter account set session policy policies.public.account_session_policy;
...
+GRANT ROLE ANALYTICS_PRD_READ TO ROLE TRANSFORMER_DEV;
 ```
 
 !!! note
-    This grant is not managed via terraform in order to keep the configurations of
-    different environments as logically separate as possible. We may revisit this
-    decision should the manual grant cause problems.
+These grants are not managed via Terraform in order to keep the configurations of
+different environments as logically separate as possible. We may revisit this
+decision should the manual grants cause problems.
```
ian-r-rose, line 146: These need to be indented for `!!! note`

ian-r-rose: Docs [here](https://python-markdown.github.io/extensions/admonition/)

ram-kishore-odi: Thank you ! I updated it.

PR #559, opened 2025-07-18
Schema validation macro

ian-r-rose:
Fixes https://github.com/cagov/caldata-infrastructure-template/issues/55.

I went through a number of different approaches to solving this problem (as an aside, GitHub copilot was actually really helpful in iterating on the different approaches, though I had to keep a pretty close eye on it!). The different approaches are listed below, with pros and cons to each of them. Ultimately, I went with the **run-operation** approach, as it didn't have any of the deal-breaker drawbacks that the other approaches had.

### Model contract

Pros:

* Officially supported by dbt
* Easy to integrate with a dbt-workflow: automatically runs with every build

Cons:

* No support for [sources or seeds](https://docs.getdbt.com/docs/mesh/govern/model-contracts#where-are-contracts-supported) (I consider this a deal-breaker)
* No way to configure project-wide. (I consider this a deal-breaker)

### post-hook

Pros:

* Possible to configure project-wide
* Integrates very well with a dbt workflow: always runs after any model build
* Relatively straightforward to implement
* Easy to drop into a project with a single file

Cons:

* Can't run on sources (cc @summer-mothwood, who flagged this problem. I consider this a deal-breaker)

### dbt tests

Pros:

* Can run on both sources and models
* Integrates very well with a dbt workflow: always runs after any model build!

Cons:

* Can't configure project-wide (I consider this a deal-breaker)

### dbt run-operation

Pros:

* Integrates well with a dbt workflow, though it doesn't run very granularly (i.e., runs project-wide rather than on the specific subset of models you run)
* Works for both sources and models
* Easy to run project-wide
* Easy to drop into a project with a single file

Cons:

* Not much. Only downside is that it doesn't automatically run for every model, it would be a separate CI check.

### Python script

Pros:

* More flexible than Jinja

Cons:

* Doesn't integrate with a dbt workflow: if someone with a dbt profile configured wanted to run it, they would also need to make sure their python environment is set up to run the script
* Some clients are not comfortable with Python
* More difficult to drop into a project


britt-allen (approved):
tried it out on your branch, this is really neat!

MACKAT05 (approved):
Looks good. i was able to run this in situ.

PR #570, opened 2025-09-05
Documented SSL error workarounds

JamesSLogan:
Documenting a few findings, per @britt-allen's suggestion. These problems aren't uncommon on corporate networks/machines, as far as I can tell from researching the errors.

I also fixed 2 small things `mkdocs` was complaining about.


JamesSLogan:
I removed the schema validation step in the CI job, as we've done in other repositories.

`dbt run-operation validate_all_schemas --args '{"errors_only": true}'`

britt-allen (approved):
ty James! Note to self to port this to training docs as well!

PR #573, opened 2025-09-23
Common policy module and Sentinel user

ram-kishore-odi:
Hello @ian-r-rose,

Can you please review the information below and the PR and let me know your feedback? I apologize for the delay in getting this functionality into the repo compared to the original plan.

Development plan
[r-plan-dev.txt](https://github.com/user-attachments/files/22503261/r-plan-dev.txt)

Production plan
[r-plan-prd.txt](https://github.com/user-attachments/files/22503263/r-plan-prd.txt)

# Mv commands for the policies

Terraform mv commands to make sure nothing gets destroyed and added 

## Database
terraform state mv 'snowflake_database.policies' \
  'module.policies.snowflake_database.policies'

## Password policy + account attachment
terraform state mv 'snowflake_password_policy.user_password_policy' \
  'module.policies.snowflake_password_policy.user_password_policy'

terraform state mv 'snowflake_account_password_policy_attachment.attachment' \
  'module.policies.snowflake_account_password_policy_attachment.attachment'

## Authentication policies
terraform state mv 'snowflake_authentication_policy.external_duo_mfa' \
  'module.policies.snowflake_authentication_policy.external_duo_mfa'

terraform state mv 'snowflake_authentication_policy.service_account_keypair' \
  'module.policies.snowflake_authentication_policy.service_account_keypair'

terraform state mv 'snowflake_authentication_policy.legacy_service_password' \
  'module.policies.snowflake_authentication_policy.legacy_service_password'

terraform state mv 'snowflake_authentication_policy.odi_okta_only[0]' \
  'module.policies.snowflake_authentication_policy.odi_okta_only[0]'

terraform state mv 'snowflake_authentication_policy.admin_okta_duo[0]' \
  'module.policies.snowflake_authentication_policy.admin_okta_duo[0]'

## Account default auth policy attachment
terraform state mv 'snowflake_account_authentication_policy_attachment.default_policy[0]' \
  'module.policies.snowflake_account_authentication_policy_attachment.default_policy[0]'

## Existing Sentinel service user and its role grant during the module refactor

terraform state mv \
  'module.elt.snowflake_legacy_service_user.sentinel' \
  'module.policies.snowflake_legacy_service_user.sentinel'

terraform state mv \
  'module.elt.snowflake_grant_account_role.logger_to_sentinel' \
  'module.policies.snowflake_grant_account_role.logger_to_sentinel'

#### Note: the new configuration sets name = "SENTINEL_SVC_USER", while the existing object is "SENTINEL_SVC_USER_PRD". If the actual Snowflake user hasn’t been renamed, Terraform will still attempt to recreate the resource after the state move. So please share your preferred option below - 

To avoid this, we have two options:

#### Option 2: Rename the user in Snowflake to SENTINEL_SVC_USER, then run terraform import so the state matches the renamed object.

Given that the BlueVoyant/CDT team has not yet responded, it may be helpful to retain the current name, even though it is not ideal and includes "PRD," indicating it is a production account. However, I am also completely fine with waiting for their feedback and then renaming it. Please let me know what you think

### DIFF
--- a/terraform/snowflake/modules/policies/main.tf
+++ b/terraform/snowflake/modules/policies/main.tf
@@ -37,37 +37,36 @@ resource "snowflake_database" "policies" {
 # account-level role & warehouse (no env vars)
 resource "snowflake_account_role" "logger" {
   provider = snowflake.securityadmin
-  name     = "LOGGER_PRD"
+  name     = "LOGGER"
   comment  = "Account-level role for logging tasks"
 }
 
 resource "snowflake_warehouse" "logging" {
   provider            = snowflake.sysadmin
-  name                = "LOGGING_PRD_WH"
+  name                = "LOGGING_WH"
   warehouse_size      = "XSMALL"
   auto_suspend        = 60
   auto_resume         = true
   initially_suspended = true
-  comment             = "Warehouse used by Sentinel/logging tasks (PRD)"
+  comment             = "Warehouse used by Sentinel/logging tasks"
 }
 
-# ßSentinel service user
-resource "snowflake_legacy_service_user" "sentinel" {
-  provider = snowflake.useradmin
-  name     = "SENTINEL_SVC_USER"
-  comment  = "Service user for Sentinel"
-
-  lifecycle { ignore_changes = [rsa_public_key] }
-
+# Sentinel service user
+resource "snowflake_service_user" "sentinel" {
+  provider          = snowflake.useradmin
+  name              = "SENTINEL_SVC_USR"
+  login_name        = "SENTINEL_SVC_USR"
   default_warehouse = snowflake_warehouse.logging.name
   default_role      = snowflake_account_role.logger.name
+  disabled          = false
+  comment           = "Sentinel service user"
 }
 
-# Grant LOGGER_PRD to Sentinel
+# Grant LOGGER to Sentinel
 resource "snowflake_grant_account_role" "logger_to_sentinel" {
   provider  = snowflake.useradmin
   role_name = snowflake_account_role.logger.name
-  user_name = snowflake_legacy_service_user.sentinel.name
+  user_name = snowflake_service_user.sentinel.name
 }
 
 # Default user password policy
@@ -128,7 +127,7 @@ resource "snowflake_authentication_policy" "external_duo_mfa" {
   mfa_authentication_methods = ["PASSWORD"]
   mfa_enrollment             = "REQUIRED"
   client_types               = ["SNOWFLAKE_UI", "DRIVERS", "SNOWSQL"]
-  comment                    = "Duo-MFA-only authentication policy for external human users (PRD)"
+  comment                    = "Duo-MFA-only authentication policy for external human users"
 }
 
 resource "snowflake_authentication_policy" "service_account_keypair" {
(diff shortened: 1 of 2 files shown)


terraform/snowflake/modules/policies/variables.tf
```diff
@@ -0,0 +1,21 @@
...
+
+variable "logging_warehouse_name" {
+  description = "The name of the warehouse to be used by the Sentinel service user."
+  type        = string
+}
+
+variable "logger_role_name" {
+  description = "The name of the role to be granted to the Sentinel service user."
+  type        = string
+}
```
ian-r-rose, line 21: I don't see a strong reason for these to be configurable (all of our other functional roles are pretty hard-coded)

terraform/snowflake/modules/policies/main.tf
```diff
@@ -0,0 +1,145 @@
...
+  schema                     = "PUBLIC"   # Schema name
+  name                       = "legacy_service_password"
+  authentication_methods     = ["PASSWORD"]
+  client_types               = ["DRIVERS", "SNOWSQL"]
+  comment                    = "Password-only authentication policy for legacy service accounts"
+}
+
+# Set odi_okta_only as the default authentication policy for the account
+resource "snowflake_account_authentication_policy_attachment" "default_policy" {
+  count = var.okta_integration_name == null ? 0 : 1
```
ian-r-rose, line 120: I suspect the `terraform fmt` check might have some problems here

terraform/snowflake/modules/policies/main.tf
```diff
@@ -0,0 +1,145 @@
...
+        snowflake.useradmin,
+      ]
+    }
+  }
+  required_version = ">= 1.0"
+}
+
+// Create the POLICIES database to store the password policy
+resource "snowflake_database" "policies" {
+  name = var.policies_database_name // name of the policies database from variable
```
ian-r-rose, line 30: This will potentially run into some problems if it is applied after the docs [here](https://cagov.github.io/data-infrastructure/setup/snowflake-setup/#apply-a-session-policy). There's a few places in the docs where we should do some updating (can be in a follow-up)

terraform/snowflake/modules/policies/main.tf
```diff
@@ -0,0 +1,145 @@
...
+  name     = "SENTINEL_SVC_USER"
+  comment  = "Service user for Sentinel"
+  lifecycle {
+    ignore_changes = [rsa_public_key]
+  }
+
+  # Use the input variable here
+  default_warehouse = var.logging_warehouse_name
+  # Use the input variable here
+  default_role      = var.logger_role_name
```
ian-r-rose, line 137: Let's just create the role and warehouse in this module

ian-r-rose (commented):
Thanks @ram-kishore-odi, this is looking good to me, a few comments about the structure, but it's looking close

ian-r-rose:
To your question about the sentinel service user: I don't think it's possible to move it due to the name change. In my opinion, it's okay to destroy and recreate it with the different name, and let CDT know. If their internal processes can't handle that, it's kind of on them.

terraform/snowflake/modules/policies/main.tf
```diff
@@ -130,16 +151,14 @@ resource "snowflake_legacy_service_user" "sentinel" {
   lifecycle {
     ignore_changes = [rsa_public_key]
   }
-
-  # Use the input variable here
-  default_warehouse = var.logging_warehouse_name
-  # Use the input variable here
-  default_role      = var.logger_role_name
+# no cross-module references; use resources created here
```
ian-r-rose, line 154: This comment doesn't seem particularly helpful, as it refers to a review conversation rather than anything in the code

terraform/snowflake/modules/policies/variables.tf
```diff
@@ -1,3 +1,9 @@
+variable "environment" {
+  description = "Environment suffix"
+  type        = string
+  default     = "PRD"
+}
```
ian-r-rose, line 5: This is meant for account-level things. Why do we have an `environment` variable?

PR #590, opened 2026-02-26
SCPRS data

ian-r-rose:
This adds a scraper and supporting (very light) dbt models for SCPRS data.

It's what's backing [this](https://coda.io/d/State-IT-Procurement_dAv5u98nxj0/State-IT-Procurement_sud-NQiZ) Coda page, allowing for up-to-date monitoring of State IT procurements


airflow/dags/procurement/scprs.py
britt-allen: this file is really long. would you consider putting the two scraper classes in separate files?

PR #618, opened 2026-07-30
dlt test pipeline for azure to snowflake

britt-allen:
This is a minimal dlt pipeline that loads two Azure SQL database tables into Snowflake. The pipeline ran successfully and the data can be found in `raw_dev.dlt`. This closes [this Asana task](https://app.asana.com/1/1202865175765955/project/1210910778445814/task/1213937737882597?focus=true).

### DIFF
--- a/.dlt/config.toml
+++ b/.dlt/config.toml
@@ -1,0 +1,6 @@
+# put your configuration values here
+
+[runtime]
+log_level="WARNING"  # the system log level of dlt
+# use the dlthub_telemetry setting to enable/disable anonymous usage data reporting, see https://dlthub.com/docs/reference/telemetry
+dlthub_telemetry = false
--- a/.gitignore
+++ b/.gitignore
@@ -131,6 +131,9 @@ ENV/
 env.bak/
 venv.bak/
 
+# dlt credentials (keep secrets out of git; config.toml is fine to track)
+.dlt/secrets.toml
+
 # Spyder project settings
 .spyderproject
 .spyproject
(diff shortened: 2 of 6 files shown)


ian-r-rose (commented):
The code for running the pipeline looks good to me (I like how short it is!).

The thing that's missing here is the orchestration piece. You ran this locally, and it worked. I can't see the way you connected to the database (that config is probably in a dlt secrets file that's uncommitted?). The question then becomes, what's next? Running something once is different from running an incremental sync several times per day.

It's a little undecided what we might want to do for a tool like dlt, though in the linked Asana task we have:

> running dlt inside of a Snowflake stored procedure could work and be pretty low infrastructure. We could also use Airflow for that.

I think we could merge this and create a follow-on task for orchestration, or could do it here. But doing the orchestration piece would force you to answer questions like:

1. What is the service account we should use here?
2. What's the authentication mechanism to the database?
3. Do we need to allow-list any additional IP addresses?
4. What frequency should we sync?

ram-kishore-odi (commented):
Are you also planning to check in the `.gitignore` file? Just checking.

britt-allen:
> Are you also planning to check in the `.gitignore` file? Just checking.

Check for what @ram-kishore-odi ? I am not sure I understand.

ram-kishore-odi:
> > Are you also planning to check in the `.gitignore` file? Just checking.
> 
> Check for what @ram-kishore-odi ? I am not sure I understand.

Hi @britt-allen, I just meant whether the .gitignore change was intentional and whether you were planning to include it in this PR as well. No issue if it is.

PR #623, opened 2026-09-10
Security bump 2026-09-09

ram-kishore-odi:
Regenerates `uv.lock` with `uv lock --upgrade` to pull in security fixes. Clears 5 of the 8 open Dependabot alerts for this repo.

Supersedes #622.

The three remaining alerts (`pymdown-extensions` and `mkdocs-material`) need `mkdocs-material` past the `~=9.6.14` pin in `pyproject.toml`, so they are out of scope for a lockfile refresh and are left for a separate change.

Locks were generated with uv 0.12.5. uv 0.9.x cannot parse the relative `exclude-newer` value and silently resolves with no cutoff, so a green `uv lock --check` on an older uv does not prove the cooldown applied. Newest `upload-time` here is 2026-09-02, inside the cutoff.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -87,30 +87,30 @@ wheels = [
 
 [[package]]
 name = "boto3"
-version = "1.43.71"
+version = "1.43.87"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "botocore" },
     { name = "jmespath" },
     { name = "s3transfer" },
 ]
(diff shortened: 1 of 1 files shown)


ram-kishore-odi:
Closing this one rather than merging it. Every GitHub check passed but the dbt Cloud run failed (run 31610), and the same lockfile-refresh PR passed dbt Cloud three weeks ago in #621, so this is not a platform outage. dbt Cloud passes on the other open PRs here, so it looks specific to this refresh.

The most likely candidate is `dbt-core-experimental-parser` moving 2.0.0b1 to 2.0.0rc1; `dbt-core` itself only moved 1.12.2 to 1.12.3. I cannot read the dbt Cloud run log to confirm, so I would rather not merge a change I cannot validate against the repo that everything else depends on.

Note that Dependabot #622 covers the `sqlparse` advisory on its own and its dbt Cloud check passes, so that alert can be cleared without this PR.

PR #624, opened 2026-09-17
version bump

britt-allen:
attempts to fix this failure: https://github.com/cagov/data-infrastructure/actions/runs/35134351849


ian-r-rose (approved):
Looks good to me, thanks @britt-allen!

